PR #367, opened 2018-02-15
Core 1081: Java 8 compatibility

harbu:
### Overview
Figured out what was causing problems with JDK 8:

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00000001026f3f8c, pid=9039, tid=0x0000000000003b03
#
# JRE version: Java(TM) SE Runtime Environment (8.0_152-b16) (build 1.8.0_152-b16)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.152-b16 mixed mode bsd-amd64 compressed oops)
# Problematic frame:
# V  [libjvm.dylib+0xf3f8c]  GraphBuilder::try_inline_full(ciMethod*, bool, Bytecodes::Code, Instruction*)+0x4d6
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
#

...

Current CompileTask:
C1:  40401 14287       3       com.unifina.utils.MapTraversal::getBoolean (121 bytes)
```

I don't know exactly why this happens but I assume it has to do with method overloading.


test/functional/LoginTesterSpec.groovy
```diff
@@ -1,10 +1,14 @@
...
-
+import geb.driver.CachingDriverFactory
 import pages.CanvasPage
 import pages.LoginPage
 import geb.spock.GebReportingSpec
 
 abstract class LoginTesterSpec extends GebReportingSpec {
+
+	def setupSpec() {
+		resetBrowser()
```
jtakalai: :D  what's the performance hit (in seconds total functional test running time)?

jtakalai:
there are a bunch of @CompileStatics I've had to comment out to make it build before. I never committed them though.

test/functional/LoginTesterSpec.groovy
harbu: I think it wasn't much. Maybe couple minutes at max? It's not done after every test, just every test file so the hit isn't that bad.

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -116,8 +116,7 @@ class StreamService {
 	}
 
 	// Ref to Kafka will be abstracted out when Feed abstraction is reworked
-	@CompileStatic
-	void sendMessage(Stream stream, @Nullable String partitionKey, long timestamp, byte[] content, byte contentType, int ttl=0) {
+	void sendMessage(Stream stream, String partitionKey, long timestamp, byte[] content, byte contentType, int ttl=0) {
```
hpihkala: @harbu Was it the `@CompileStatic` that was causing the crash? This is one of the most frequently called methods, would be nice to have that performance benefit there. If it's a no-go, then.. no can do.

harbu: Not necessarily caused by `@CompileStatic`, but probably due to overloading of method name.

I can add the `@CompileStatic` back in and run all tests when I have a chance.

kare: I've tried it now and its looking good!

kare: First I ran all tests once without the CompileStatic annotation patch and no SIGSEGV.
Second I've run all tests twice with the CompileStatic annotation patch above and they passed without SIGSEGV!

mikhaelsantos:
Once this is done I can update Travis to run the tests with java8!

PR #371, opened 2018-02-27
Core 1185: Ethereum identity uniqueness

harbu:
- Add database-indexed field `idInService` to `IntegrationKey`. Represents
public identifier of an integration key in the context of the service.
This is equivalent to public wallet address for services ETHEREUM and
ETHEREUM_ID.
- Enforce the uniqueness of public addresses of `IntegrationKey`(s) of type `ETHEREUM_ID` at service level.

### DIFF
--- a/grails-app/domain/com/unifina/domain/security/IntegrationKey.groovy
+++ b/grails-app/domain/com/unifina/domain/security/IntegrationKey.groovy
@@ -20,7 +20,8 @@ class IntegrationKey implements Serializable {
 	static mapping = {
 		id generator: IdGenerator.name // Note: doesn't apply in unit tests
 		json type: 'text'
-		idInService(index: "id_in_service_idx")
+		idInService(index: "id_in_service_and_service_idx")
+		service(index: "id_in_service_and_service_idx")
 	}
 
 	static constraints = {
--- a/grails-app/domain/com/unifina/domain/security/IntegrationKey.groovy
+++ b/grails-app/domain/com/unifina/domain/security/IntegrationKey.groovy
@@ -12,6 +12,7 @@ class IntegrationKey implements Serializable {
 	String name
 	Service service
 	String json
+	String idInService
 
 	Date dateCreated
 	Date lastUpdated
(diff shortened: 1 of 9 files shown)


grails-app/domain/com/unifina/domain/security/IntegrationKey.groovy
```diff
@@ -12,13 +12,15 @@ class IntegrationKey implements Serializable {
...
 	String json
+	String idInService
 
 	Date dateCreated
 	Date lastUpdated
 
 	static mapping = {
 		id generator: IdGenerator.name // Note: doesn't apply in unit tests
 		json type: 'text'
+		idInService(index: "id_in_service_idx")
```
hpihkala: Maybe have both fields (service, idInService) in the index?

grails-app/services/com/unifina/service/EthereumIntegrationKeyService.groovy
```diff
@@ -81,24 +83,28 @@ class EthereumIntegrationKeyService {
...
 		try {
 			byte[] messageHash = ECRecover.calculateMessageHash(message)
 			address = ECRecover.recoverAddress(messageHash, signature)
 		} catch (SignatureException | DecoderException e) {
 			throw new ApiException(400, "ADDRESS_RECOVERY_ERROR", e.message)
 		}
-		IntegrationKey key = new IntegrationKey(
+
+		if (IntegrationKey.findByServiceAndIdInService(IntegrationKey.Service.ETHEREUM_ID, address) != null) {
+			throw new DuplicateNotAllowedException("Ethereum ID already taken.")
```
hpihkala: How about: "This Ethereum address is already associated with another Streamr user".

harbu:
I read about this and there is an argument against and for.

A multicolumn index `(a, b)` can utilize the first column `a` without the presence of 2nd column `b` in the query (but not vice versa.). So putting `service` before `id_in_service` would allow for the use of the index, for example, in `SELECT * FROM integration_key WHERE service = 'ETHEREUM'`. However I didn't see such queries in our codebase. All cases in which we search by `service`, we also search by `user`, which already cuts down the search results a lot. 

The benefit of putting `id_in_service` before `service` is that it is a more selective column (has more values) which apparently has performance benefits in narrowing down the number of results sooner.

Overall, if we foresee that we will be searching through field `service` without the help of other indexes it may be worthwhile turning the columns of the index the other way around. I have a hunch that putting `id_in_service` before `service` might be a  micro-optimization of sorts.

PR #372, opened 2018-02-28
Feature/byte array decoding modules

yaruno:
Generic byte array decode modules. String to byte array and byte array to string. Needed for decoding Lorawan packet payloads.


src/java/com/unifina/signalpath/text/DecodeByteArrayToString.java
```diff
@@ -0,0 +1,63 @@
...
+	private StringOutput error = new StringOutput(this, "error");
+	private DecodeParameter decodeType = new DecodeParameter(this, "decodeType", DecodeParameter.HEX);
+
+	@Override
+	public void init() {
+		addInput(in);
+		addInput(decodeType);
+		addOutput(out);
+		addOutput(error);
+		in.setCanConnect(true);
```
harbu: Type `Input` does not have method `setCanConnect(boolean)`?

src/java/com/unifina/signalpath/text/DecodeByteArrayToString.java
```diff
@@ -0,0 +1,63 @@
...
+		decodeType.setCanConnect(false);
+		decodeType.setUpdateOnChange(true);
```
harbu: I think it should work without `setUpdateOnChange = true`.

src/java/com/unifina/signalpath/text/DecodeStringToByteArray.java
```diff
@@ -0,0 +1,61 @@
...
+			error.send("Failed to parse : '" + in.getValue() + "'");
+		}
+	}
+
+	@Override
+	public void clearState() {
+
+	}
+
+	public static class DecodeParameter extends StringParameter {
```
harbu: Same class as in `DecodeByteArrayToString`? Could be extracted from inner class to its own class and then re-used in both classes to avoid repetition.

src/java/com/unifina/signalpath/text/DecodeByteArrayToString.java
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	@Override
+	public void sendOutput() {
+
+		try {
+			switch (decodeType.getValue()){
+				case DecodeParameter.BASE64: out.send(DatatypeConverter.printBase64Binary(in.getValue()));
+					break;
```
harbu: Funky indentation

src/java/com/unifina/signalpath/text/DecodeByteArrayToString.java
```diff
@@ -0,0 +1,63 @@
...
+				default:
+				case DecodeParameter.HEX: out.send(DatatypeConverter.printHexBinary(in.getValue()));
+					break;
+			}
+		} catch (Exception e ){
+			error.send("Failed to parse byte array");
```
harbu: Can a more specific Exception be caught? If not, would be good idea to log error, i.e.,

`private static final Logger log = Logger.getLogger(DecodeByteArrayToString.class);`

and then

`log.error(e);`

src/java/com/unifina/signalpath/text/DecodeStringToByteArray.java
```diff
@@ -0,0 +1,61 @@
...
+		try {
+			switch (decodeType.getValue()){
+				case DecodeParameter.HEX: out.send(DatatypeConverter.parseHexBinary(in.getValue()));
+					break;
+				default:
+				case DecodeParameter.BASE64: out.send(DatatypeConverter.parseBase64Binary(in.getValue()));
+					break;
+			}
+		} catch (Exception e ){
+			error.send("Failed to parse : '" + in.getValue() + "'");
```
harbu: Can a more specific Exception be caught? If not, would be good idea to log error, i.e.,

`private static final Logger log = Logger.getLogger(DecodeStringToByteArray.class);`

and then

`log.error(e);`

test/unit/com/unifina/signalpath/text/DecodeByteArrayToStringSpec.groovy
```diff
@@ -0,0 +1,48 @@
...
+	def module
+
+	def setup() {
+		module = null
+	}
+
+	def cleanup() {
+
+	}
+
```
harbu: Can get rid of everything above, just use `module` as a local variable instead.

test/unit/com/unifina/signalpath/text/DecodeStringToByteArraySpec.groovy
```diff
@@ -0,0 +1,56 @@
...
+	def module
+
+	def setup() {
+		module = null
+	}
+
+	def cleanup() {
+
+	}
+
```
harbu: Can get rid of everything above, just use `module` as a local variable instead.

grails-app/migrations/core/2018-02-28-bytearray-decoding-modules.groovy
```diff
@@ -0,0 +1,29 @@
...
+			column(name: "category_id", valueNumeric: 27)
+			column(name: "implementing_class", value: "com.unifina.signalpath.text.DecodeStringToByteArray")
+			column(name: "name", value: "DecodeStringToByteArray")
+			column(name: "js_module", value: "GenericModule")
+			column(name: "type", value: "module")
+			column(name: "module_package_id", valueNumeric: 1)
+			column(name: "json_help", value: '')
+		}
+	}
+	changeSet(author: "jarno", id: "decode-byte-array-to-string") {
```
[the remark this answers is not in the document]
yaruno: ah, now they are unique decode-string-to-byte-array & decode-byte-array-to-string but if you prefer some other format it can be changed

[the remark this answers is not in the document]
harbu: ah nvm, I first misread that they were the same.

src/java/com/unifina/signalpath/text/DecodeByteArrayToString.java
jtakalai: it does

src/java/com/unifina/signalpath/text/DecodeStringToByteArray.java
jtakalai: they might be separate later if there's different features in coding directions available.

PR #375, opened 2018-02-28
change state parameter from running to RUNNING

Aapzu:


web-app/react-app/actions/canvas.js
```diff
@@ -17,7 +17,7 @@ export const getRunningCanvases = () => (dispatch: Function) => {
     dispatch(getCanvasesRequest())
     return axios.get(createLink(apiUrl), {
         params: {
-            state: 'running',
+            state: 'RUNNING',
```
hpihkala: Case should follow what comes out of the API. So for example if you get a canvas and you get `{ ... state: 'running' }`, then the same should work for filtering.

hpihkala: Atm Grails seems to bind `CanvasListParams#state` enum automatically, and that's why it only works in upper case. In `Canvas.State` there is a `fromValue` implementation but that doesn't get used. 

Would the whole upper case vs. lower case mess be simplified if the enum name is changed from `State.RUNNING` to `State.running`? @harbu thoughts?

harbu: API (i.e. `Canvas.toMap()`) returns uppercase value so this makes it more consistent.

PR #376, opened 2018-02-28
CORE-1172 webpack bundles no longer handled by grails resources plugin

hpihkala:
- Webpack bundles are now served without interference from Grails resources plugin
- No more problems with Grails not detecting changed resources
- In dev mode, bundles are named `bundleName.bundle.js`
- In prod mode, bundles are named `bundleName.bundle.[hash].js`
- For Grails to know the hashed filenames, some magic was added to `_Events.groovy` which persists those war-build-time filenames to `application.properties`. Those properties automatically appear in `grailsApplication.metadata`.
- `WebpackTagLib` has tags for including the webpack resources in a way that abstracts the complexity of detecting the correct filename.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,6 @@ node_modules
 .slcache/
 plugin.xml
 hs_err_pid*
-web-app/js/unifina/webpack-bundles/*
+web-app/webpack-bundles/*
 npm-debug.log*
 tomcat.8081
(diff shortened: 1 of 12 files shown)


Aapzu:
@hpihkala how do you want to proceed with this? Can I merge? I'd like to do a little addition but I can do it as another PR

PR #379, opened 2018-03-02
Core 1112 Ethereum identity UI

kare:


web-app/react-app/components/ProfilePage/IdentityHandler/web3.js
```diff
@@ -0,0 +1,13 @@
+// @flow
```
Aapzu: Just noticed that this script doesn't notice if the account is changed (or metamask is unlocked). Maybe this is something @jtakalai could help to develop?
Either it could notice it through some event (and change the defaultAccount) or we need to use the function `ownWeb3.eth.getAccounts()` in the actions (I think that even `ownWeb3.eth.accounts` doesn't update on account change

web-app/react-app/components/ProfilePage/IntegrationKeyHandler/IntegrationKeyHandlerSegment/index.jsx
```diff
@@ -74,7 +78,7 @@ export class IntegrationKeyHandlerSegment extends Component<Props> {
                     />
                     <IntegrationKeyHandlerInput
                         fields={this.props.inputFields}
-                        onNew={this.onNew}
+                        onNew={this.props.onNew || this.onNew}
```
Aapzu: Getting the `onNew` function as a prop is the right way to go👍🏻 (we/I need to change it also in other parts of the code)

web-app/react-app/actions/identity.js
```diff
@@ -0,0 +1,89 @@
...
+        .then((signature) => {
+            const input = {
+                ...integrationKey,
+                signature: signature,
+                address: ownWeb3.eth.defaultAccount,
+                challenge: {
+                    ...challenge
+                },
+            }
+            return axios.post(createLink(apiUrl), input)
```
Aapzu: Could we use `createIntegrationKey` from `actions/integrationKey.js` here? (Now the identities are removed with `deleteIntegrationKey` action so it would follow the line). it would then go somehow like this:
```
return dispatch(createIntegrationKey(input))
    .then(...)
```
The only problem is the error handler (which now triggers a notification so then there would be two notifications). Maybe I'll think about this a while and then we could have a call about it?

web-app/react-app/reducers/integrationKey.js
```diff
@@ -29,6 +30,7 @@ const initialState = {
 export default function(state: IntegrationKeyState = initialState, action: IntegrationKeyAction): IntegrationKeyState {
 
     switch (action.type) {
+        case CREATE_IDENTITY_REQUEST:
```
Aapzu: I really think that the reducer files should correspond the action files. So then the actions considering the identity creation should be in `reducers/identity.js`

web-app/react-app/components/ProfilePage/IdentityHandler/web3.js
jtakalai: There's a hack for it, but the official docs say poll with setTimeout  /:

web-app/react-app/components/ProfilePage/IdentityHandler/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+    render() {
+        return (
+            <Panel header="Identities">
+                <Row>
+                    <IntegrationKeyHandlerSegment
+                        onNew={this.onNew}
+                        service={this.props.service}
+                        name="Ethereum"
+                        inputFields={[]}
+                        tableFields={[]}
```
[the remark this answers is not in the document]
kare: If I remove inputFields I'll get a runtime error.

web-app/react-app/actions/identity.js
```diff
@@ -0,0 +1,89 @@
...
+                })
+
+        })
+        .catch((response) => {
+            const err = parseError(response)
+            dispatch(createIdentityFailure(err))
+            dispatch(error({
+                title: 'Create identity failed',
+                message: err.message,
+                autoDismiss: 0,
```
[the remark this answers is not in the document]
kare: I thought that user would be better informed in error case. I'll remove those and follow the common style.

web-app/react-app/actions/identity.js
kare: If we try to reuse the code from createIntegrationKey() it leads to more complexity. So I don't think that a little duplication here hurts at all?

web-app/react-app/reducers/integrationKey.js
kare: Reducers and actions now reside in integrationkey.js

web-app/react-app/components/ProfilePage/IdentityHandler/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+
+    onDelete = (id: $ElementType<IntegrationKey, 'id'>) => {
+        this.props.deleteIntegrationKey(id)
+    }
+
+    render() {
+        const hasWeb3 = getWeb3().isEnabled()
+        return (
+            <Panel header="Ethereum Identities">
+                <p key={0}>
```
mondoreale: Why do we `key` it? It's not an array, is it?

Aapzu: Oh, my bad. It was an array for a (very little) while and forgot them...

PR #386, opened 2018-03-20
Core 1187 data balance

kare:


test/unit/com/unifina/service/BalanceServiceSpec.groovy
```diff
@@ -0,0 +1,50 @@
...
+import spock.lang.Specification
+
+import java.util.concurrent.ExecutionException
+
+@TestFor(BalanceService)
+class BalanceServiceSpec extends Specification {
+	def address = "0x494531425508c4Bc95E522b24fd571461583E916"
+	def setup() {
+		service.web3 = Mock(Web3Balance)
+    }
```
harbu: Indentation off

kare: I think that is just GitHub UI. Actual code looks good to me!

kare:
Web3j has it's own tests. Should we rely on those or re-test it?

harbu:
Yeah I agree we shouldn't re-test what is already unit tested in a 3rd party library. I was thinking more of the lines of testing that the service is properly configured in a real environment. But maybe that would be best tested by a end-to-end test that exercises the eventual REST API endpoint that will use this service.

.idea/engine-and-editor-grailsPlugins.iml
```diff
@@ -307,6 +307,18 @@
           <root url="jar://$MAVEN_REPOSITORY$/org/apache/ant/ant-launcher/1.9.6/ant-launcher-1.9.6.jar!/" />
           <root url="jar://$MAVEN_REPOSITORY$/org/apache/ant/ant/1.9.6/ant-1.9.6.jar!/" />
           <root url="jar://$MAVEN_REPOSITORY$/cglib/cglib/3.2.6/cglib-3.2.6.jar!/" />
+          <root url="jar://$MAVEN_REPOSITORY$/com/squareup/okhttp3/logging-interceptor/3.8.1/logging-interceptor-3.8.1.jar!/" />
```
jtakalai: whoa, which of these are for testing?

jtakalai: or maybe they're just web3j deps or something? This is a project file, I just noticed.

grails-app/services/com/unifina/service/BalanceService.groovy
```diff
@@ -0,0 +1,58 @@
...
+
+import javax.annotation.PostConstruct
+import java.util.concurrent.ExecutionException
+
+class BalanceService {
+	protected Web3Balance web3
+
+	@PostConstruct
+	void init() {
+		final HttpService httpService = new HttpService("https://rinkeby.infura.io") // TODO: Fix address
```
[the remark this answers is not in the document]
jtakalai: infura is fine I guess, though I think we'd want some amount of redundancy...

another thing is maybe using websockets instead of http, but for querying balances... whatever is fine

PR #393, opened 2018-03-27
CORE-1186: Product score 

harbu:
Products will be ordered (by default) by score.


grails-app/domain/com/unifina/domain/marketplace/Product.groovy
```diff
@@ -18,6 +18,7 @@ class Product {
 
 	Date dateCreated
 	Date lastUpdated
+	Integer score = 0
```
kare: Please add a short comment that explains what and why of score

kare:
@harbu LGTM!

PR #394, opened 2018-03-28
CORE-534, CORE-1057: Ask before leaving canvas if unsaved changes

jtakalai:
where unsaved changes currently is approximated with "module added". But it's better than nothing.

### DIFF
--- a/grails-app/views/canvas/editor.gsp
+++ b/grails-app/views/canvas/editor.gsp
@@ -23,6 +23,8 @@
 	var loadBrowser
 	var saveAndAskName
 
+	var hasUnsavedChanges = false
+
 $(function() {
 	$('#moduleTree').bind('loaded.jstree', function() {
 		Tour.startableTours([0])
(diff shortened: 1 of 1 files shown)


Aapzu:
This doesn't really catch the changes. Only if modules are added through the search.
I've already started this in [CORE-580-canvas-confirmation-on-exit](https://github.com/streamr-dev/engine-and-editor/tree/CORE-580-canvas-confirmation-on-exit), please take a look at that

jtakalai:
Good point, all functional tests probably break. Eeeeeh. Ok, drop this change.

BUT: we're already ditching bug reports because "editor refactor will make them obsolete". Now we can't deploy a fix because functional tests for editor will break. But then again, those tests will be obsolete when the editor refactor comes?

I'm not suggesting tossing the functional tests here and now, but I feel it's a bit silly situation that a fix to this problem by Aapeli (in CORE-580-canvas-confirmation-on-exit branch) just sits there because fixing tests is hard...

harbu:
I think this should be fixed/developed (semantics: I see this more as an improvement than as a bug), but we just need to make sure that functional tests are passing and fixed along with the ride. Fixing them should be relatively "copy-pasty" once it is figured out how to press "ok" on the alert dialog with Selenium / Geb.

PR #399, opened 2018-04-05
Core 1240 thumbnail

kare:


src/java/com/unifina/utils/ImageResizer.java
```diff
@@ -0,0 +1,48 @@
...
+			this.width = width;
+			this.height = height;
+		}
+		public int width() { return width; }
+		public int height() { return height; }
+	}
+
+	public byte[] resize(byte[] image, String filename, Size size) throws IOException {
+		final BufferedImage original = ImageIO.read(new ByteArrayInputStream(image));
+		final Image tmp = original.getScaledInstance(size.width(), size.height(), Image.SCALE_SMOOTH);
```
jtakalai: Cropping from center would probably be better than stretching (ping @juhah), that is, something like
```
final float aspectRatio = (float)size.width() / (float)size.height();
final int w = original.getWidth();
final int h = original.getHeight();
final BufferedImage cropped = w > h * aspectRatio ?
    original.getSubImage((w - h * aspectRatio) / 2, 0, h * aspectRatio, h) : 
    original.getSubImage(0, (h - w / aspectRatio) / 2, w, w / aspectRatio);
final Image tmp = cropped.getScaledInstance(size.width(), size.height(), Image.SCALE_SMOOTH);
```

jtakalai: plus missing int casts here and there probably  (:

juhah: Both work for me, altough centering would probably be easier on the eyes 😎

jtakalai: made a new ticket about that: https://streamr.atlassian.net/browse/CORE-1265 just in case that becomes interesting in the future.

src/java/com/unifina/utils/ImageResizer.java
```diff
@@ -0,0 +1,48 @@
...
+		g.dispose();
+
+		final ByteArrayOutputStream baos = new ByteArrayOutputStream();
+		ImageIO.write(resized, guessImageFormat(filename), baos);
+		return baos.toByteArray();
+	}
+
+	private String guessImageFormat(final String filename) {
+		if (filename == null || filename.length() == 0) {
+			return "jpg";
```
harbu: Why jpg as default? Should we just throw error instead?

test/unit/com/unifina/controller/api/ProductApiControllerSpec.groovy
```diff
@@ -424,7 +424,7 @@ class ProductApiControllerSpec extends Specification {
 			controller.uploadImage()
 		}
 		then:
-		1 * productImageService.replaceImage(product, bytes)
+		1 * productImageService.replaceImage(product, bytes, _)
```
harbu: Would be good to check last parameter here "my-product-image.jpg"

grails-app/services/com/unifina/service/ProductImageService.groovy
```diff
@@ -3,20 +3,43 @@ package com.unifina.service
...
 	IdGenerator idGenerator = new IdGenerator()
+	ImageResizer imageResizer = new ImageResizer()
 
-	void replaceImage(Product product, byte[] fileBytes) {
+	void replaceImage(Product product, byte[] fileBytes, String filename) {
 		imageVerifier.verifyImage(fileBytes)
-		String newImageUrl = fileUploadProvider.uploadFile("product-images/${idGenerator.generate()}", fileBytes)
+		String newImageUrl;
+		String newThumbnailUrl;
+		for (ImageResizer.Size size : ImageResizer.Size.values()) {
```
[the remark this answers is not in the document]
kare: Initial idea was to use enum for easy addition of new image formats and compiler check for switch (Turns out I didn't remember that old java compiler doesn't know about such checks).

I'll change this to more readable format.

PR #407, opened 2018-04-10
Core 1260 add owner to product

kare:


scripts/db-diff.sh
```diff
@@ -16,3 +16,7 @@ grails dbm-gorm-diff $FILENAME --add $2 --stacktrace
 # Add package to beginning of file
 { printf "package core\n"; cat grails-app/migrations/$FILENAME; } > grails-app/migrations/$FILENAME.new
 mv grails-app/migrations/$FILENAME{.new,}
+
+sed -i '' -e '/^$/d' grails-app/migrations/changelog.groovy
+sed -i '' -e '/^$/d' -e 's/[[:space:]](generated)//' "grails-app/migrations/$FILENAME"
```
harbu: Can you add a comment to code explaining what does this does?

PR #412, opened 2018-04-11
Findbugs

kare:
Review super carefully!

### DIFF
--- a/src/java/com/unifina/feed/AbstractFeedProxy.java
+++ b/src/java/com/unifina/feed/AbstractFeedProxy.java
@@ -1,16 +1,15 @@
 package com.unifina.feed;
 
-import java.lang.reflect.InvocationTargetException;
-import java.util.Queue;
-import java.util.concurrent.ConcurrentLinkedQueue;
-
+import com.unifina.data.FeedEvent;
 import com.unifina.data.IEventRecipient;
+import com.unifina.domain.data.Feed;
+import com.unifina.utils.Globals;
 import grails.util.Holders;
 import org.apache.log4j.Logger;
 
-import com.unifina.data.FeedEvent;
-import com.unifina.domain.data.Feed;
-import com.unifina.utils.Globals;
+import java.lang.reflect.InvocationTargetException;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
 
 
 /**
@@ -165,7 +164,7 @@ public abstract class AbstractFeedProxy<ModuleClass, RawMessageClass, MessageCla
 
 			// TODO: remove debug
 			if (firstRealQueue == null && checkAge) {
-				log.info("First real time message: " + counter + ". Events: " + events);
+				log.info("First real time message: " + counter + ".");
 				firstRealQueue = counter;
 			}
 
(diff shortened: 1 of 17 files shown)


src/java/com/unifina/signalpath/ModuleOption.java
```diff
@@ -125,7 +123,9 @@ public static boolean validate(Object option) {
 				List<Map<String, String>> possibleValues = (List<Map<String, String>>) opt.get("possibleValues");
 				boolean found = false;
 				for (Map<String, String> value : possibleValues) {
-					if (value.get("value") == opt.get("value") || value.get("value").equals(opt.get("value"))) {
+					final String val = value.get("value");
+					final Object optValue = opt.get("value");
+					if (val.equals(optValue)) {
```
harbu: Can we add `val != null && ` infront?

jtakalai: semantics are changed by this patch. Definitely add a null check.

Notice though that `opt.get("value")` is unchanged inside the loop, so it should be taken outside. That whole inner loop thing should probably be something like
```
String selected = (String)opt.get("value");
return possibleValues.stream().anyMatch(v -> ((String)v.get("value")).equals(selected));
```

src/java/com/unifina/feed/AbstractFeedProxy.java
```diff
@@ -165,7 +164,7 @@ private void processAndQueue(long counter, MessageClass msg, boolean checkAge) {
 
 			// TODO: remove debug
 			if (firstRealQueue == null && checkAge) {
-				log.info("First real time message: " + counter + ". Events: " + events);
```
jtakalai: changed this into this branch

src/java/com/unifina/security/UserClassLoader.java
```diff
@@ -76,35 +63,7 @@ protected ModuleClassCollector(InnerLoader cl, CompilationUnit unit, SourceUnit
 			super(cl,unit,su);
 		}
     }
-
```
jtakalai: if no test breaks, it's probably dead and good to go (this looks kind of fundamental, so either it's useful and everything breaks; or not). But static analysis might not catch stuff that's called by
* groovy magic
* reflection through fully qualified class name stored in MySQL (modules)

Of course both of those could be cool to get rid of eventually...

kare: it was reported as unused by IntelliJ too. Unit tests pass on my machine.

src/java/com/unifina/feed/AbstractFeedProxy.java
```diff
@@ -164,7 +164,7 @@ private void processAndQueue(long counter, MessageClass msg, boolean checkAge) {
 
 			// TODO: remove debug
 			if (firstRealQueue == null && checkAge) {
-				log.info("First real time message: " + counter + ".");
+				log.info("First real time message: " + counter + ". Event count: " + events.length);
```
kare: Potential NPE?

src/java/com/unifina/feed/AbstractFeedProxy.java
harbu: ?

src/java/com/unifina/feed/AbstractFeedProxy.java
harbu: Apparently yes, since there is a null check later on in the same code section. (Though pretty bad practice to allow null to be returned by the `process` method.)

harbu: @jtakalai

kare: I simply removed whole log call based on TODO comment above.

PR #414, opened 2018-04-11
CORE-1257: API endpoint for publishing free products

harbu:
API endpoint `POST /api/v1/products/:productId/deployFree` for deploying free Products (price = 0).

## Details
- New controller method `ProductApiController#deployFree`
- New service `FreeProductService` and method `deployFreeProduct`
- Bring back switch-statement to fix issue with proxy/assist domain classes in `PermissionService#getResourcePropertyName`
- rest e2e tests for endpoint
- swagger for endpoint
- unit tests for service and controller

### DIFF
--- a/grails-app/services/com/unifina/service/PermissionService.groovy
+++ b/grails-app/services/com/unifina/service/PermissionService.groovy
@@ -1,11 +1,17 @@
 package com.unifina.service
 
 import com.unifina.api.NotPermittedException
+import com.unifina.domain.dashboard.Dashboard
+import com.unifina.domain.data.Feed
+import com.unifina.domain.data.Stream
+import com.unifina.domain.marketplace.Product
 import com.unifina.domain.security.Key
 import com.unifina.domain.security.Permission
 import com.unifina.domain.security.Permission.Operation
 import com.unifina.domain.security.SecUser
 import com.unifina.domain.security.SignupInvite
+import com.unifina.domain.signalpath.Canvas
+import com.unifina.domain.signalpath.ModulePackage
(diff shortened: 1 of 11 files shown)


grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -482,9 +488,22 @@ class PermissionService {
...
+			return "feed"
+		} else if (resource instanceof ModulePackage) {
+			return "modulePackage"
+		} else if (resource instanceof Product) {
+			return "product"
+		} else if (resource instanceof Stream) {
+			return "stream"
+		} else {
+			throw new IllegalArgumentException("Unexpected resource class: " + resource)
+		}
```
kare: First feeling about this is that perhaps it could be simplified by just calling DomainClass.getResourcePropertyName() and let domain classes define their own resource property names. But I'm not sure if it actually simplifies anything?

Common base class implementation won't probably help here since its getClass() returns the base class name.

It would be a great idea to have a dynamically generated resource name because it would be refactor safe, but it might be too much work due to proxy assist stuff.

harbu: I think one nice solution would be to have an interface `Permissionable` (or something better) that a domain class must implement to be able to use the Permission system. The `Permissionable` interface would require you implement method `getResourcePropertyName`. This would also grant type-safety.

harbu: Made JIRA ticket out of this. See CORE-1292

PR #418, opened 2018-04-13
Core 1297 javascript injections

jtakalai:
Mainly please review the first commit (b827).

Additional suggestions for escapings are welcome, though maybe better merge this thing first.

### DIFF
--- a/grails-app/views/canvas/editor.gsp
+++ b/grails-app/views/canvas/editor.gsp
@@ -162,7 +162,7 @@ $(function() {
 
 	$(SignalPath).on('saved', function(event, savedJson) {
 		$('#modal-spinner').hide()
-		Streamr.showSuccess('${message(code:"signalpath.saved")}: '+savedJson.name)
+		Streamr.showSuccess('${message(code:"signalpath.saved")}: '+Streamr.escape(savedJson.name))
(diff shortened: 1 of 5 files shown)


Aapzu:
I may not fully understand the reasons behind this, but I'm struggling to get what's the idea with escaping these particular places? None of these strings are evaluated anywhere?

Aapzu:
Äh okay I didn't realize that Pnotify doesn't escape them, now I understand this fully:)

jtakalai:
Well, not just pnotify... the real problem was search autocomplete, but also the "do you want to save" dialogs.

jtakalai:
This again was a bit of patching; the "real" solution (I imagine) would be to not push <script> stuff into DB in the first place; and more systematically sanitize user input. But this patches that particular case (<script>s in stream, canvas names, descriptions)

jtakalai:
grails already does the escaping so stuff like https://www.streamr.com/canvas/editor/5furFq2hRza8fyxzXxxgnQlZXMOWm3TMKng43VyYCEaw and https://www.streamr.com/stream/show/-uSTUpa1Summ0roiwhrM0A isn't problematic in the list/show views

PR #420, opened 2018-04-16
CORE-1232: add missing callback

harbu:
Added callback to `subscriptionService.afterProductUpdated` so that stream permissions get updated after a stream is removed or added to a Product.


harbu:
Merging this branch, package-lock.json thing should be fixed in branch CORE-1160-marketplace-api.

PR #424, opened 2018-04-19
CORE-1230: Remove troll user's products

kare:


harbu:
Just landed CORE-1160-marketplace-api to master.  Can you base this PR onto master?

kare:
Done!

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -13,6 +13,20 @@ class ProductService {
 	PermissionService permissionService
 	SubscriptionService subscriptionService
 
+	void removeUsersProducts(String username) {
+		def query = Product.where {
```
harbu: I think can be shortened to `def all = Product.findAllByOwner(username)`

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -13,6 +13,20 @@ class ProductService {
...
+			owner == username
+		}
+		def all = query.findAll()
+		all.each{ it ->
```
harbu: `all.each { Product product ->` and can get rid of cast below

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -13,6 +13,20 @@ class ProductService {
...
+			Product product = (Product) it
+			product.streams.toArray().each{ Stream stream ->
```
harbu: Are you sure the streams need to be deleted manually? I think there is a GORM feature that will auto-delete associations?

grails-app/controllers/com/unifina/controller/api/RemoveUsersProductsController.groovy
```diff
@@ -0,0 +1,20 @@
...
+import com.unifina.service.ProductService
+import grails.plugin.springsecurity.annotation.Secured
+
+@Secured(["IS_AUTHENTICATED_ANONYMOUSLY"])
+class RemoveUsersProductsController {
+	ProductService productService
+
+	@StreamrApi
+	def index() {
+		if (params.username == null) {
```
harbu: We need to check that logged-in user `request.apiUser` is an admin. There is a `SecRole` called `ROLE_ADMIN` that indicates that a user is an admin. Take a look at `SecUser#isDevOps` to see how to check for a role.

harbu: Another option would be to add a new level authentication level to `@StreamrApi`, but I think I will make a separate JIRA task out of that for now.

grails-app/services/com/unifina/service/ProductService.groovy
kare: I wanted to make sure that callback in removeStreamFromProduct() gets called, but I'll drop it.

grails-app/services/com/unifina/service/ProductService.groovy
kare: Thats what I tried initially, but its not working as one would expect!
```
Compilation error: startup failed:
/engine-and-editor/grails-app/services/com/unifina/service/ProductService.groovy: 18: Expected parameter of type java.lang.Object but got com.unifina.domain.marketplace.Product
 @ line 18, column 13.
   		all.each{ Product product ->
               ^

/engine-and-editor/grails-app/services/com/unifina/service/ProductService.groovy: 19: [Static type checking] - Cannot find matching method java.lang.Object#delete(java.util.LinkedHashMap <java.lang.String, java.lang.Boolean>). Please check if the declared type is right and if the method exists.
 @ line 19, column 4.
   			product.delete(flush: true)
      ^

2 errors

```

kare: I think my grails environment is just messed

kare: Well both, got my grails env cleaned and setup, but still same error as above

kare:
I wonder if I missed removing something?

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -13,6 +14,28 @@ class ProductService {
 	PermissionService permissionService
 	SubscriptionService subscriptionService
 
+	void removeUsersProducts(String username) {
+		def all = Product.findAllByOwner(username)
+		all.toArray().each{ Product product ->
+			product.delete(flush: true)
```
harbu: I think this needs to be done last because `flush: true` will attempt to perform the SQL deletion immediately and will fail because there are foreign keys still linked to it.

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -13,6 +14,28 @@ class ProductService {
...
+			// Delete users all subscriptions
+			def user = SecUser.where {
+				username == username
```
harbu: The idea is not to delete the troll user's subscriptions but rather all the subscriptions to the troll user's Products.

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -13,6 +14,28 @@ class ProductService {
...
+			}.find()
+			def subscriptions = subscriptionService.getSubscriptionsOfUser(user)
+			subscriptions.toArray().each { Subscription subscription ->
+				subscriptionService.deletePermissions(subscription)
+				subscription.delete(flush: true)
+			}
+			// Remove all streams from product
+			if (product.streams != null) {
```
harbu: Product.streams should always be non-null.

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -13,6 +14,28 @@ class ProductService {
...
+				product.streams.toArray().each{ Stream stream ->
+					removeStreamFromProduct(product, stream)
```
harbu: I think the best order to do things here (in pseudo-code) would be:
```
for product in listOfTrollUsersProducts:
  for subscription in product.subscriptions:
     subscription.delete() // => should delete permissions, if not, loop through permissions and delete
  for stream in product.streams:
      product.removeFromStreams(stream)
  product.delete(flush: true)
```

kare:
It currently runs these statements:
Hibernate: delete from product_streams where product_id=?
Hibernate: delete from subscription where id=? and version=?

Hibernate: delete from product where id=? and version=?
Hibernate: delete from product_streams where product_id=?
Hibernate: delete from product where id=? and version=?

Test has two products. Other product has subscription.

grails-app/services/com/unifina/service/SubscriptionService.groovy
```diff
@@ -14,6 +14,16 @@ import grails.compiler.GrailsCompileStatic
 class SubscriptionService {
 	PermissionService permissionService
 
+	void deleteProduct(Product product) {
```
harbu: Can we have a more accurate name, e.g. `deleteAllSubscriptionsOfProduct(Product product)` or `deleteAllByProduct(Product product)` something along those lines.

grails-app/services/com/unifina/service/SubscriptionService.groovy
```diff
@@ -14,6 +14,16 @@ import grails.compiler.GrailsCompileStatic
...
+		def user = SecUser.where {
+			username == product.owner
+		}.find()
+		getSubscriptionsOfUser(user).toArray().each { Subscription subscription ->
```
harbu: If I understand correctly, this is deleting the subscriptions of the "troll user".  What I think we should be doing, is deleting subscriptions of the Product.

So instead of `getSubscriptionsOfUser(user)`, something like
`List<Subscription> subscriptions = Subscription.findAllByProduct(product)` ?

PR #429, opened 2018-04-26
CORE-1348: pricePerSecond as String

harbu:
- `Product#toMap` returns `pricePerSecond` as a String instead of a Long
-  Apparently prices can be sent in JSON body as either String or Long, both will be accepted. Thus no changes required on that front.


hpihkala:
This will break things in the front end, which is not expecting `pricePerSecond` to be a string (used to be a number).  Should not merge this before @Aapzu gives a go ahead.

hpihkala:
@Aapzu's `BigNumber` change is now in marketplace master. Should be ok to land this now. Watch out for price-related bugs in front-end.

PR #431, opened 2018-04-26
CORE-1330: Make API and user guide documentation public

harbu:
Make API and user guide documentation public.

Tested by hand that
- Works when not logged in
- Works when logged in (and API key is included in Swagger)


harbu:
You can set the API key when not logged in via the Swagger UI. If not set no `Authorization` header will be sent.

PR #438, opened 2018-04-30
CORE-1345: Convert Product.owner String to SecUser

kare:


grails-app/domain/com/unifina/domain/marketplace/Product.groovy
```diff
@@ -100,7 +101,7 @@ class Product {
 			pricePerSecond: pricePerSecond,
 			priceCurrency: priceCurrency.toString(),
 			minimumSubscriptionInSeconds: minimumSubscriptionInSeconds,
-			owner: owner
+			owner: owner.name
```
hpihkala: This will lead to N+1 db requests when listing products. I think @harbu had a way to make GORM eager-fetch `owner`. It would be a major improvement in terms of performance.

scripts/db-diff.sh
```diff
@@ -19,4 +19,6 @@ mv grails-app/migrations/$FILENAME{.new,}
 
 sed -i '' -e '/^$/d' grails-app/migrations/changelog.groovy # remove empty lines
 sed -i '' -e '/^$/d' -e 's/[[:space:]](generated)//' "grails-app/migrations/$FILENAME" # replace " (generated)" with ""
+# replace generated numeric (len 13-14) changeset id with $1
```
hpihkala: Nice! +1

grails-app/domain/com/unifina/domain/marketplace/Product.groovy
harbu: Added optimization!

PR #439, opened 2018-05-01
CORE-1357 remove captcha from signup by default

hpihkala:
Captcha on signup can now be enabled/disabled via a configuration option. It is now **disabled** by default. The reason is that we are using Google reCaptcha and that is blocked by the Chinese firewall, so Chinese can't sign up without VPN.


hpihkala:
Ran func test `RegisterSpec` manually, passing

PR #440, opened 2018-05-01
CORE-1282 always use ajax login and redirect client-side

hpihkala:
- Fixes CORE-1282 (reopened) and CORE-1366 (new)
- Login form will now always submit credentials as an ajax request and handle post-login redirection client-side
- `redirect` query param is passed to the login view by a custom `authenticationEntryPoint` bean (implemented by class `RedirectAppendingAuthenticationEntryPoint`)
- `Login*Spec` integration tests are passing

Originally started looking into this due to the post-login redirection not working in prod. Turns out it was a configuration issue: the marketplace was using `https://streamr.com/marketplace/...` as redirect target, and it was not passing the validation of `LoginRedirectValidator`. I made it slightly more permissible and added logging to avoid encountering this kind of "mystery bugs" in the future.

Anyway, due to CORE-1366 I started to research this alternative login approach and ended up with this proposal, which should work for all present and future use cases.


src/java/com/unifina/security/RedirectAppendingAuthenticationEntryPoint.java
```diff
@@ -0,0 +1,66 @@
...
+		String url = super.buildRedirectUrlToLoginPage(request, response, authException);
+
+		// If we are looking for the base url or the default redirect, don't add the redirect query param
+		if (request.getRequestURI().equals(getFullURI("/"))
+				|| request.getRequestURI().equals(getFullURI(defaultRedirectURI))) {
+			return url;
+		} else if (LoginRedirectValidator.isValid(request.getRequestURL().toString())){
+			return url + "?redirect="+request.getRequestURL();
+		} else {
+			log.info("Redirect url rejected: "+request.getRequestURL());
```
harbu: Good idea to log these!

test/unit/com/unifina/controller/security/LoginRedirectValidatorSpec.groovy
```diff
@@ -19,10 +19,15 @@ class LoginRedirectValidatorSpec extends Specification {
...
+		true   | "https://marketplace.streamr.com/sub"
+		true   | "https://streamr.com"
+		true   | "https://streamr.com/"
+		true   | "https://streamr.com/marketplace"
 		false  | "http://www.streamr.com/"
-		false  | "https://streamr.com/"
+		false  | ""
 		false  | null
 		false  | "https://www.google.fi/"
 		false  | "https://www.streamr.com.phissing.com/"
```
harbu: Add test case for `https://www.streamr.com.phissing.com`

PR #442, opened 2018-05-02
CORE-1224: Related Products API endpoint

kare:


grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -12,6 +12,30 @@ class ProductService {
...
+
+	List<Product> relatedProducts(String id) {
+		Product product = Product.get(id)
+		if (product == null) {
+			return new ArrayList<Product>()
+		}
+		// find Product.owner's other products
+		Set<Product> all = new HashSet<Product>(Product.findAllByOwner(product.owner))
+		// find other products from the same category
+		Set<Product> cat = new HashSet<Product>(Product.findAllByCategory(product.category))
```
hpihkala: - Result set size could be hard-limited to some sensible max (e.g. 20)
- Unauthenticated fetch - may return products which the user does not have access to! Easiest solution is probably to use `ProductService#list` and pass the category in `ProductListParams`

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -12,6 +12,30 @@ class ProductService {
...
+		all.addAll(cat)
+
+		final int max = 3 // max number of related products
```
hpihkala: Give as query param, with some default value (e.g. 3)? Should have an enforced maximum (e.g. 10)

test/unit/com/unifina/service/RelatedProductsSpec.groovy
```diff
@@ -0,0 +1,105 @@
...
+		//
+		// results should not list:
+		//   - p1 (product used to find related products)
+		//   - p4 (different category and user)
+		//   - p6 (different category and user)
+
+		// UI is specified to show three (3) related products
+	}
+
+	void "find related products"() {
```
hpihkala: Is the branch where `all.size() > max` being tested (the one where random elements are removed from the set)? To test with randomness, you need to to set a seed in the test for `Random` (or mock it) so that it will produce deterministic results.

PR #445, opened 2018-05-02
CORE-1346: Duplicate results returned by API listing endpoints

harbu:
### Summary
Addresses issue in which a resource can appear several times in the JSON response list produced by an API endpoint.

### Details
If a user has two or more permissions with the same access-level (e.g. ` Operation.READ`) to a resource, the resulting list of resources produced by `apiService#list` and `permissionService#createUserPermissionCriteria` will contain that resource multiple times. The desire is that a resource appear at maximum once per results (even across pagination).

This was fixed by modifying the database query that Grails produces. By grouping the `permission` table by resource id (e.g. `canvas_id`, `dashboard_id`, `stream_id` etc.) we ensure that a resource appear at most once in the result set before joining with the resource database table. 

### Concerning tests
Unfortunately changes had to be made to multiple tests because of spotty test support of GORM.  Specifically, the `groupProperty` function has not been stubbed in the GORM test support library thereby causing runtime errors for all tests that attempt to use `permissionService#createUserPermissionCriteria`. To summarize:

- Multiple tests from `PermissionServiceSpec` were moved to `PermissionServiceIntegrationSpec`
- Elsewhere `PermissionService` was mocked to allow tests to pass in unit test environment.

### DIFF
--- a/grails-app/services/com/unifina/service/PermissionService.groovy
+++ b/grails-app/services/com/unifina/service/PermissionService.groovy
@@ -164,7 +164,7 @@ class PermissionService {
 			return []
 		}
 
-		Closure permissionCriteria = createUserPermissionCriteria(userish, op, includeAnonymous)
+		Closure permissionCriteria = createUserPermissionCriteria(resourceClass, userish, op, includeAnonymous)
 
 		return resourceClass.withCriteria {
 			permissionCriteria.delegate = delegate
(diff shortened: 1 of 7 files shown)


kare:
It bothers me a lot that these nice commit summaries get lost after merge. They should be part of the commit message. Another thing is that in my opinion we should be using rebase instead of squash and merge. It would focus work a lot better and your commit messages wouldn't disappear between merges with master.

PR #449, opened 2018-05-04
CORE-1378: fix infinite loop in related products

harbu:
- Fix inifinite loop in `ProductService#relatedProducts`, see comment in "Files changed".
- Respond with `Product#toSummaryMap()` instead of `Product`.
- Add `publicAccess=true` to fetches inside `ProductService#relatedProducts` to make sure that results are returned.
- Some refactorings

### DIFF
--- a/grails-app/services/com/unifina/service/ProductService.groovy
+++ b/grails-app/services/com/unifina/service/ProductService.groovy
@@ -17,27 +17,24 @@ class ProductService {
 	Random random = ThreadLocalRandom.current()
 
 	List<Product> relatedProducts(Product product, int maxResults, SecUser user) {
-		if (product == null) {
-			return new ArrayList<Product>()
-		}
 		// find Product.owner's other products
-		ListParams params = new ProductListParams(productOwner: product.owner, max: maxResults)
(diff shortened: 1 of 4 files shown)


grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -17,27 +17,24 @@ class ProductService {
...
+		params = new ProductListParams(categories: [product.category].toSet(), max: maxResults, publicAccess: true)
+		all.addAll(list(params, user))
 
-		all.removeIf { Product p -> p.id == product.id }
-		if (all.size() <= maxResults) {
-			return new ArrayList<Product>(all)
-		}
-		while (all.size() > maxResults) {
-			int i = random.nextInt(all.size() - 1)
-			all.remove(i)
```
harbu: This isn't removing by index, but instead trying to remove `Integer i` from the Set. Thus the set never decreases in size.

kare: 👍  I was thinking about a list. just missed the conversion from set to list

PR #452, opened 2018-05-11
CORE-1226: query params addModule and addStream to /canvas/editor

jtakalai:
Currently has an issue (on my very local dev env at least) that the correct canvas only flashes by, is replaced with empty canvas. Might have something to do with the tour being active.

Henri marked as reviewer mostly to comment on why request.JSON is in canvas/editor controller code (it should be a GET, without body, so what should request.JSON be and when?)

### DIFF
--- a/grails-app/controllers/com/unifina/controller/core/signalpath/CanvasController.groovy
+++ b/grails-app/controllers/com/unifina/controller/core/signalpath/CanvasController.groovy
@@ -48,9 +48,6 @@ class CanvasController {
 		def endDate = new Date()
 		def currentUser = SecUser.get(springSecurityService.currentUser.id)
 
-		// TODO: what's the case for request.JSON in editor action? GETs don't have bodies. Where can a POST come from? What can it possibly do with the response?
-		def json = request.JSON
-
 		[
 			beginDate: beginDate,
 			endDate: endDate,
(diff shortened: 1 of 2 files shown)


harbu:
So currently when user hasn't completed Tour, the canvas will wiped out? Could we prevent the tour from activating when either of these parameters have been supplied? 

Regarding the `request.JSON`, looks quite weird indeed. Maybe there is some code path that does a `POST` or `PUT` on that controller method? Or maybe it is just old code?

grails-app/controllers/com/unifina/controller/core/signalpath/CanvasController.groovy
```diff
@@ -47,6 +47,8 @@ class CanvasController {
 		def beginDate = new Date()
 		def endDate = new Date()
 		def currentUser = SecUser.get(springSecurityService.currentUser.id)
+
+		// TODO: what's the case for request.JSON in editor action? GETs don't have bodies. Where can a POST come from? What can it possibly do with the response?
```
hpihkala: It's probably some old remnant from AlgoCanvas times, maybe once used via a `POST` request. I don't think it's used anymore. It can be removed.

harbu:
I added in prevention of Tour from starting when `addModule` or `addStream` url parameters are given.

PR #454, opened 2018-05-21
[CORE-1410] Stream messages need to be escaped before showing in UI

Aapzu:


Aapzu:
@jtakalai what do you mean? I don't think that anything that is not written by us should be rendered as HTML in the DOM

PR #455, opened 2018-05-22
[CORE-1411] Use replace instead of assign to redirect in login form

Aapzu:


Aapzu:
@kare have we decided something about commit message policy? I agree that the message should be descriptive but it also shouldn't be more than 50 characters long. Your suggestion is 79 characters long.
There's also no need to describe the reason for the commit. The JIRA issue id in the message already tells all things needed.

kare:
I'd like to have both "what" and "why" covered in commit message for future reference.

Your commit message does great on the "what" part.

There is a good write up about writing commit messages (just mentally replace CL with PR):
https://groups.google.com/forum/#!msg/golang-dev/6M4dmZWpFaI/SyU5Sl4zZLYJ

> "It's likely that a day or month or a year from now someone will search the history log for the change but not be able to find it because all the important information is in the code but not identified in the CL description."

Searching for an old commit might be hard. It will be even harder if you have to cross reference some of the information from an external system.

Aapzu:
@kare good points although I don't agree with them. But reasoning my side is irrelevant now, since they're both just opinions. 

I looked at quite many commit messages in e&e, and could only find one that answered to the questions "why". Squash merge might strip the comment bodies off, but anyway in master this was the case. Also don't remember anyone talking about this in the company ever. Therefore it'd be unreasonable to call this a company policy (or even a clear code principle). 

You'd clearly like to make it a company policy. This happens by raising conversation about it on general level, and then by deciding together if we want to make it so. Not by demanding it to happen just because you decided so. 

Until then I'm not going to change the message. It tells clearly what has been done and goes along the earlier messages. If you have no other comments, I'm going to consider this as Approved. 

P.S. You can still use whatever message you like, I don't mind

PR #456, opened 2018-05-23
[CORE-1413] Should not be able to addStream without a permission

Aapzu:


grails-app/controllers/com/unifina/controller/core/signalpath/CanvasController.groovy
```diff
@@ -47,6 +50,17 @@ class CanvasController {
 		def beginDate = new Date()
 		def endDate = new Date()
 		def currentUser = SecUser.get(springSecurityService.currentUser.id)
+		def addStream
+		def error
+
+		if (params.addStream) {
```
jtakalai: Come to think of it; yes, this needs to be solved in server-side. However, shouldn't it be impossible to get the unauthorized stream module in the first place? I mean, it's fine that there's an error when page with addStream is loaded; but shouldn't the error really appear when it tries to get that module from backend; that is, in ModuleController.jsonGetModule?

That check could be done to modules too (if that module exists or whatever), although something like https://www.streamr.com/canvas/editor?addModule=3453 gives an error (not the best one, but still a decent hint)

jtakalai: like, does this change fix that ModuleController.jsonGetModule might still give that module if requested directly.

It's a bit of code smell that there's a `// TODO: check that the owning user has the privileges to access this module` in ModuleService.getModuleInstance... the rabbit hole probably goes pretty deep simply because when a module is requested it is in fact instantiated and then just JSON is extracted. Basically the place where stream permission should be checked is ConfigurableStreamModule.onConfiguration, I guess. It should throw if the apiUser doesn't have read permission to that stream; although of course there will be no apiUser when the module is instantiated when a canvas is simply run.

jtakalai: There's also AbstractStreamSourceModule.getStream, which could also be a cool place for the permission check. Now the only thing here is, where should that user whose permissions are to be checked come from...

jtakalai: I think @harbu and/or @hpihkala should comment on this.

jtakalai: I think this is one of those places where engine and editor are kind of a bit too intertwined; so far this kind of permission thing hasn't been a problem, but it will increasingly be.

Aapzu:
This was just the first fix to hide the horrible bug that allows you to add any stream to a canvas with addStream and see it's content without any permissions

grails-app/controllers/com/unifina/controller/core/signalpath/CanvasController.groovy
hpihkala: > However, shouldn't it be impossible to get the unauthorized stream module in the first place?

Not always, it's a bit more complicated than that. Example situation: someone shares a Canvas with you. The Canvas contains a Stream module, so you need to be able to get the module (to view the Canvas) even though you don't have access to the Stream directly.

hpihkala: Permission checking as suggested by @jtakalai has been tried in the past, and that approach unfortunately has unintended side effects such as the one above.

Adding a module with particular settings can't really be prevented (for example, anyone can craft a JSON that contains a module trying to load an unauthorised Stream). The damage from Stream modules appearing on Canvases is limited to leaking Stream metadata (name, fields) to someone who knows the ID. Higher priority should be assigned preventing unauthorised access to the data itself in the stream.

PR #459, opened 2018-05-29
CORE-1406: Improve error message for email already in use

harbu:


grails-app/services/com/unifina/service/UserService.groovy
```diff
@@ -27,8 +28,17 @@ class UserService {
 		user.enabled = true
 
 		if (!user.validate()) {
-			log.warn(checkErrors(user.errors.getAllErrors()))
-			throw new UserCreationFailedException("Registration user validation failed: " + checkErrors(user.errors.getAllErrors()))
+			def errors = checkErrors(user.errors.getAllErrors())
+			log.warn(errors)
```
kare: Is this log necessary? In the long run I'd like to cut down noise from the logs.

harbu: I think registration occurs somewhat rarely that it is useful to have the logging here still in case something goes wrong. Though I do agree that we need to cut down noise in logging in general.

PR #460, opened 2018-05-31
CORE-1332: add role checking to StreamrApi-annotation

harbu:
`@StreamrApi` annotation now supports role requirements. By default everything works as before (no roles are required.) 

Usage:
```
@StreamrApi(authenticationLevel = AuthLevel.USER, allowRoles = AllowRole.DEVOPS)
```

or with multiple roles (of which at least one must be satisfied by logged-in user)

```
@StreamrApi(authenticationLevel = AuthLevel.USER, allowRoles = [AllowRole.DEVOPS, AllowRole.ADMIN])
```

Note that `authenticationFilter` has not been (unit) tested in this PR. The reason is that in Grails there are no tools to really unit test a filter by itself, but rather this is done in the context of writing a unit test for a controller that uses the filter. We should make sure that the functionality gets tested when writing first controller that uses role requirements with the annotation.


harbu:
Yeah they're the Spring Security roles that are checked. What I am building next (admin tools for viewing and managing running canvases) will use this annotation.

PR #461, opened 2018-05-31
CORE-1414: Should not be able to start a canvas if it has a Stream without permission

jtakalai:
now checked SHALLOWLY upon starting. Can still be circumvented by taking it as subcanvas. These problems should vanish once data points are checked against stream permissions.

### DIFF
--- a/grails-app/services/com/unifina/service/SignalPathService.groovy
+++ b/grails-app/services/com/unifina/service/SignalPathService.groovy
@@ -4,11 +4,13 @@ import com.google.gson.Gson
 import com.unifina.api.CanvasCommunicationException
 import com.unifina.datasource.IStartListener
 import com.unifina.datasource.IStopListener
+import com.unifina.domain.data.Stream
 import com.unifina.domain.security.Permission
 import com.unifina.domain.security.SecUser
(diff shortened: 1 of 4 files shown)


grails-app/services/com/unifina/service/SignalPathService.groovy
```diff
@@ -101,6 +103,15 @@ class SignalPathService {
 			sp = (SignalPath) serializationService.deserialize(canvas.serialization.bytes)
 		}
 
+		// require read access to all streams when starting
+		// can be problematic when collaborating on shared canvas; though even then it makes sense to force
+		//   explicit read rights sharing to streams on that canvas
+		for (Stream s in sp.getStreams()) {
```
harbu: I get the feeling that this should be encapsulated inside the `SignalPath` or `SignalPathRunner` class.

jtakalai: yea it's a bit special-casing stream modules over other modules; what about modules that also use streams but are not stream modules? Now there's only SendToStream, and it does the authentication check in the module. Maybe it can be moved there if there comes lots of other modules. It would then have to be a startlistener and do that check in there.

PR #462, opened 2018-06-05
CORE-1407: Allow admin users to ping all canvases

kare:


grails-app/services/com/unifina/service/SignalPathService.groovy
```diff
@@ -239,7 +239,12 @@ class SignalPathService {
 
 		// All runtime requests require at least read permission
 		CanvasService canvasService = Holders.getApplicationContext().getBean(CanvasService) // Cannot use dependency injection because of circular dependency! Do not turn into instance variable.
-		Canvas canvas = canvasService.authorizedGetById(pathReader.readCanvasId(), user, Permission.Operation.READ)
+		Canvas canvas
+		if (user.isAdmin()) {
```
harbu: In the future we may want to look into checking for admin role inside `PermissionService`. But for now, I think this is good since we don't want the admin checking necessarily everywhere in our code base.

PR #466, opened 2018-06-21
Fix memory leak issue

harbu:
- remove (unused) catch-up functionality from `AbstractFeedProxy`
- add hard limit to the size of `DataSourceEventQueue#queue` as a precaution to avoid the queue from growing without bound. 
- make sure that the `RedisFeed` of a crashed `SignalPath` gets properly cleaned up so as to avoid leaking memory.

### DIFF
--- a/src/java/com/unifina/datasource/DataSourceEventQueue.java
+++ b/src/java/com/unifina/datasource/DataSourceEventQueue.java
@@ -3,6 +3,7 @@ package com.unifina.datasource;
 import com.unifina.data.FeedEvent;
 import com.unifina.feed.MasterClock;
 import com.unifina.utils.Globals;
+import org.apache.log4j.Logger;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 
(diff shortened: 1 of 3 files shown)


src/java/com/unifina/feed/ICatchupFeed.java
```diff
@@ -11,15 +11,6 @@
  * the normal event queue.
  */
 public interface ICatchupFeed {
```
hpihkala: This interface could be removed, I think? (related to removed catchup feature)

Also its implementation in `AbstractFeedProxy` is then unnecessary.

src/java/com/unifina/feed/AbstractFeedProxy.java
```diff
@@ -30,18 +28,10 @@
...
-	}
-
 	private static final Logger log = Logger.getLogger(AbstractFeedProxy.class);
 
 	private final MessageHub<RawMessageClass, MessageClass, KeyClass> hub;
-	private final Queue<MessageClass> realtimeWaitQueue = new ConcurrentLinkedQueue<>();
-	private final Queue<Long> realtimeWaitQueueCounter = new ConcurrentLinkedQueue<>();
 
-	private CatchupState catchupState = CatchupState.CATCHUP; // start in catchup state
 	private Catchup catchup = null;
```
hpihkala: This variable is related to the Catchup feature and could be removed. (Maybe the whole `Catchup` class can be deleted`?)

harbu:
Separated catch-up cleaning stuff to other branch. Landing this reduced change set now.

PR #477, opened 2018-07-20
[CORE-1450] Signup won't let you signup with emails with some TLDs

Aapzu:


src/groovy/com/unifina/utils/EmailValidator.groovy
```diff
@@ -0,0 +1,10 @@
+package com.unifina.utils
```
Aapzu: Extracted these to their own pr #477

PR #482, opened 2018-08-01
CORE-1469 Add Canvas.startedBy

kare:
- startedBy is set to user who started the canvas.
    - If startedBy is null, then the canvas has never been started.

### DIFF
--- a/grails-app/services/com/unifina/service/CanvasService.groovy
+++ b/grails-app/services/com/unifina/service/CanvasService.groovy
@@ -97,6 +97,7 @@ class CanvasService {
 
 		Map signalPathContext = canvas.toMap().settings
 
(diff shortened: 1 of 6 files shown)


grails-app/services/com/unifina/service/CanvasService.groovy
```diff
@@ -97,6 +97,7 @@ class CanvasService {
 
 		Map signalPathContext = canvas.toMap().settings
 
+		canvas.startedBy = asUser
```
harbu: Can this be moved into `SignalPathService#startLocal`? Feels like it belongs there since there is a bunch of other similar statements there that update the contents of the Canvas domain object.

PR #491, opened 2018-08-13
fix: CORE-1504 - Update SDKMAN:

mikhaelsantos:
* Update command for SDKMAN


mikhaelsantos:
The problem wasn't the SDK it was travis.

PR #492, opened 2018-08-15
fix: CORE-1509 - Docker image building for EE failing due to alpine-p…

mikhaelsantos:
* Update alpine-pkg-glibc from 2..27 -> 2.28
* Fix Public signing key location

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,7 @@ RUN apk add --no-cache \
             && apk add --no-cache --virtual .build-deps \
                 binutils-gold \
                 curl \
+                wget \
                 g++ \
                 gcc \
                 gnupg \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -11,6 +11,7 @@ RUN apk add --no-cache \
             && apk add --no-cache --virtual .build-deps \
                 binutils-gold \
                 curl \
+                wget \
```
kare: Just use curl instead of wget since its already there

Dockerfile
```diff
@@ -19,9 +20,12 @@ RUN apk add --no-cache \
                 make \
                 python
 
-RUN wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://raw.githubusercontent.com/sgerrand/alpine-pkg-glibc/master/sgerrand.rsa.pub
-RUN wget https://github.com/sgerrand/alpine-pkg-glibc/releases/download/2.27-r0/glibc-2.27-r0.apk
-RUN apk add glibc-2.27-r0.apk
+#
+ENV ALPINE_PKG_GLIBC_VERSION 2.28-r0
+
+RUN wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://alpine-pkgs.sgerrand.com/sgerrand.rsa.pub
```
kare: sgerrand.com tls certificate has expired. Any other alternatives?

kare: So the certificate for https://alpine-pkgs.sgerrand.com is ok, but not ok for sgerrand.com

mikhaelsantos: Checked https://alpine-pkgs.sgerrand.com which is ok. Pinged developer. Lets keep it for now.

PR #497, opened 2018-08-24
CORE-1510: application-level metrics

harbu:
Todo:
- serialization of canvases and calculate sizes

Questions:
- Any other metrics we want to gather? I think the JVM-level and server-level metrics can be gathered with other tools.

### DIFF
--- a/grails-app/controllers/com/unifina/controller/api/MetricsApiController.groovy
+++ b/grails-app/controllers/com/unifina/controller/api/MetricsApiController.groovy
@@ -21,9 +21,12 @@ class MetricsApiController {
 				running: metricsService.numOfRunningCanvases(),
 				totalSerializedBytes: metricsService.bytesUsedByRunningCanvases(),
 			],
-			eventProcessing: [
-				eventsPerSecond: eventProcessingMetrics.eventsPerSecond,
-				avgProcessingDelay: eventProcessingMetrics.avgProcessingDelay,
+			realTime: [
+				eventsPerSecond: eventProcessingMetrics.realTimeEventsPerSecond,
+				avgProcessingDelay: eventProcessingMetrics.realTimeAvgProcessingDelay,
+			],
+			historical: [
+			  eventsPerSecond: eventProcessingMetrics.historicalEventsPerSecond
 			],
 			numOfTomcatSessions: metricsService.numOfSessionsTomcat()
 		] as JSON)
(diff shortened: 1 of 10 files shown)


grails-app/controllers/com/unifina/controller/api/MetricsApiController.groovy
```diff
@@ -0,0 +1,31 @@
...
+
+
+import com.unifina.security.StreamrApi
+import com.unifina.service.MetricsService
+import grails.compiler.GrailsCompileStatic
+import grails.converters.JSON
+import grails.plugin.springsecurity.annotation.Secured
+
+@GrailsCompileStatic
+@Secured(["IS_AUTHENTICATED_ANONYMOUSLY"])
```
hpihkala: Could this be only for admin roles?

grails-app/services/com/unifina/service/MetricsService.groovy
```diff
@@ -0,0 +1,46 @@
...
+@GrailsCompileStatic
+class MetricsService {
+	SignalPathService signalPathService
+
+	int numOfRunningCanvases() {
+		signalPathService.runningSignalPaths.size()
+	}
+
+	int bytesUsedByRunningCanvases() {
+		return 0 // TODO
```
hpihkala: How about this TODO?

src/java/com/unifina/data/HistoricalEventQueue.java
```diff
@@ -154,6 +154,11 @@ public boolean process(FeedEvent event) {
 		return true;
 	}
 
+	@Override
+	protected EventQueueMetrics retrieveMetricsAndReset() {
+		return new EventQueueMetrics();
```
hpihkala: Historical runs don't contribute to metrics? (delay metric doesn't make sense, otherwise should be applicable)

grails-app/services/com/unifina/service/MetricsService.groovy
harbu: The problem is that we cannot serialize a canvas while it is running because it involves nulling fields etc. We could turn on the serialization of canvases during regular intervals, but that would affect their real-time performance guarantees.

harbu: I suggest we leave this like this for now and work on it later so that we get at least some metrics online.

hpihkala: Ok, please file JIRA ticket to follow up on this.

Are you worried about concurrency issues? The collection for this metric could happen in the canvas event queue (similarly to the normal serialization at regular intervals), yes?

PR #498, opened 2018-08-27
fix: Fix codedeploy errors:

mikhaelsantos:
* Move all codedeploy files to their own dir
* Fetch executables from codedeploy file and not AWS


.codedeploy/after_instal.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
```
harbu: Type in filename

mikhaelsantos: fixed

PR #505, opened 2018-09-04
CORE-1507: API endpoints for CSV upload

harbu:
- Grabbed CSV upload endpoints from  CORE-1075
- Refactored CSV upload logic to service-layer (see `CsvUploadService`)
- Wrote tests for endpoints


kare:
Branch name is wrong "CORE-1075-backend-changes"!
Instead of writing the uploaded file to a tmp file, should it be buffered with java.io streams for performance reasons? What is the maximum file size for CSV upload?

harbu:
I don't know if there is a good way to re-name the branch without creating a new PR. I'll just make sure to squash-and-merge with proper CORE-XXXX identifier and delete the original branch.

Buffered would mean it stays in memory, correct? I'm a bit hesitant to do that because the files may be up to 512mb-1gb of size and should be kept in memory for some time to ensure that they are there for the subsequent API calls.

tumppi:
How about scalability? If we have multiple API servers and we save the CSV file on machine that handles the request, how are we making sure that the `confirmCsvUpload` call is going to happen on the same server? Or do have sticky sessions (based on IP) configured on load balancer level?

harbu:
Yeah we have sticky sessions based on IP on load balancer level at the moment.

kare:
Quite the opposite! Here is some good information about input/output streams:
https://stackoverflow.com/questions/3010065/servlet-file-upload-memory-consumption

We could chain these calls with PipedOutput/InputStream's for multiple file upload handlers (validate upload stream/save data stream/etc).

Piped output/input streams also enable/require threading for maximum performance.
https://stackoverflow.com/questions/484119/why-doesnt-more-java-code-use-pipedinputstream-pipedoutputstream
http://www.java2s.com/Code/Java/File-Input-Output/TestPipedInputStreamandPipedOutputStreamwithThread.htm

harbu:
Unless there is a way to keep that stream (and underlying file) open between requests?

PR #510, opened 2018-09-19
CORE-1408 repair endpoint

kare:


grails-app/controllers/com/unifina/controller/api/ClusterApiController.groovy
```diff
@@ -17,45 +17,40 @@ class ClusterApiController {
 	static allowedMethods = [
 		index: "GET",
 		shutdown: "POST",
+		repair: "POST",
 	]
 
 	GrailsApplication grailsApplication
-	StreamrClient client
+	ClusterService service
```
hpihkala: This variable needs to be called `clusterService`, otherwise the automatic dependency injection won't work, and it will just be `null`! @kare did it work for you?

grails-app/services/com/unifina/service/ClusterService.groovy
```diff
@@ -0,0 +1,57 @@
+package com.unifina.service
+
+import com.streamr.api.client.CanvasesPerNode
+import com.streamr.api.client.StreamrClient
+import com.unifina.domain.signalpath.Canvas
+import groovy.transform.CompileStatic
+
+class ClusterService {
+	CanvasService canvasService
+	StreamrClient client
```
hpihkala: Same here, bean won't be injected unless name matches the bean name `streamrClient`

hpihkala:
@kare something wrong with this branch, shows a lot of unrelated changes and commits from others. Difficult to review. Can that be fixed?

hpihkala:
Tested manually, seemed to work ok!

hpihkala:
Note conflicts

PR #511, opened 2018-09-21
WIP: CORE-1445 clientside timestamps and remove SecUser.timezone

kare:


hpihkala:
@kare Quick comments:

- Chart and Table modules (possibly others) are formatting or modifying timestamps server-side. We should change the behavior to send milliseconds timestamps to the widgets and interpret them in the front end. Especially with Chart there's a hack in use, originally to make historical timestamps correct during daylight saving time. We should go over this on the phone.

- About modules that format timestamps server-side: For each, it should be considered whether we add the timezone as a parameter or option. You already did this for TimeOfDay module, but for some modules you just replaced the user timezone with UTC. Perhaps other modules need the timezone option too? Also, the timezone selection would ideally be a dropdown (with the long list of options stored frontend-side, not in canvas json) for better usability.

I'm happy to provide more detailed feedback later this week, but you might already take a look at the above points?

web-app/js/unifina/streamr-table/streamr-table.js
```diff
@@ -62,30 +62,43 @@ StreamrTable.prototype.setHeaders = function(headers) {
...
-	}
-	this.tableBody[op](newRow);
+    })
+    for (var i = 0; i < row.length; i++) {
+        let content
+        if (row[i] == null) {
+            content = ''
+        } else {
+            if (row[i] instanceof Object) {
+                if (row[i].date && row[i].streamr_date) {
```
hpihkala: Where is `date` and/or `streamr_date` written? How come there are no changes in `VariadicEventTable.java` - you're still sending server-formatted strings from there!

**Here's how to get this on track:**

- Change this line to `if (row[i].__streamr_date) {` (lower probability of name collision with user objects)
- In `VariadicEventTable` line 48 (and line 39 in `EventTable`) add to the row a `Map` containing one item with key `__streamr_date` and the timestamp in milliseconds as value for that key.
```
Map<String, Object> dateObject = new HashMap<>();
dateObject.put("__streamr_date", getGlobals().time.getTime());
nr.add(dateObject);
```
- Remove the parsing, `content = new Date(row[i].__streamr_date)` should be enough

web-app/js/unifina/test/streamr-table/streamr-table-test.js
```diff
@@ -85,6 +85,19 @@ describe('streamr-table', function() {
...
+
+		it('should replace the contents when message field has custom date object', function() {
+			table.receiveResponse({
+				id: 123,
+				nr: [
+					{ date: '2018-12-13T10:17:48.126Z' },
+					{ date: '2018-12-15T15:15:00.123Z' }
+				]
+			})
+			assert($($($parent.find('table tbody tr')[0]).find('td')[0]).text() == '2018-12-13T10:17:48.126Z')
```
hpihkala: Note that these should be formatted with `moment` to user local timezone. (see comment in `streamr-table.js`).

hpihkala:
@mondoreale Could you mind contributing to this branch the removal of asking the user timezone during the registration process?

mondoreale:
@hpihkala sure thing, sir. Will address it tomorrow real quick! 😃

mondoreale:
@hpihkala @kare in 1eddfd2 I remove timezone dropdown from the RegisterPage component along with associated js code, fixed functional tests for registration (`RegisterSpec`), and cleaned up after myself.

FYI, there's one failing test (at both 9568e09 and 1eddfd2) when I run `npm test`:

screenshot 2019-01-15 10 00 41

Cheers!

hpihkala:
@mondoreale Thanks for noticing: js tests were failing, but on Travis it looked like everything is fine because the `npm run test` script swallowed the exit code of the first command: 

`npm run test-old; npm run test-new` 

instead of 

`npm run test-old && npm run test-new`

Fixed in `master` and merged here.

hpihkala:
@kare This looks good otherwise, but there is still some JS test failing when you run `npm run test`. Could you take a look at it? It may be unrelated to these changes.

kare:
@hpihkala both javascript test suites pass on my local machine.

PR #512, opened 2018-09-28
CORE-1543 unify domain

kare:


hpihkala:
@kare same problem here, branch is corrupt with unrelated commits. Can't review. What happened? How did you add them to this branch?

grails-app/services/com/unifina/service/CanvasService.groovy
```diff
@@ -46,6 +46,9 @@ class CanvasService {
 
 	@CompileStatic
 	void updateExisting(Canvas canvas, SaveCanvasCommand command, SecUser user, boolean resetUi = false) {
+		if (command.name == null || command.name.trim() == "") {
```
hpihkala: Does it work if the default `name` is in `SaveCanvasCommand`, just like for domain objects? 
Could the code be more DRY, so that the default name is not repeated in multiple places? Something like:

In `SaveCanvasCommand`:

`String name = Canvas.DEFAULT_NAME`

In `Canvas`:

`public static String DEFAULT_NAME = "Untitled Canvas"`
...
`String name = Canvas.DEFAULT_NAME`

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -64,6 +64,9 @@ class ProductService {
 
 	Product create(CreateProductCommand command, SecUser currentUser)
 			throws ValidationException, NotPermittedException {
+		if (command.name == null || command.name.trim() == "") {
```
hpihkala: Similar comment here

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -65,6 +65,9 @@ class StreamService {
 		Stream stream = new Stream(params)
 		stream.id = id
 		stream.config = params.config
+		if (stream.name == null || stream.name.trim() == "") {
```
hpihkala: Default name is set in `Stream` already (is this code redundant?)

src/groovy/com/unifina/api/CreateProductCommand.groovy
```diff
@@ -23,6 +23,6 @@ class CreateProductCommand {
 	Long minimumSubscriptionInSeconds = 0
 
 	static constraints = {
-		importFrom(Product)
+		importFrom(Product, exclude: ["name"])
```
hpihkala: If the command object defines a default name, then can this `exclude` be removed?

If the `exclude` is needed, then is it needed in `SaveCanvasCommand` too?

grails-app/services/com/unifina/service/StreamService.groovy
kare: params variable might override it. There is a test for it:  "createStream replaces empty name with default value"

grails-app/services/com/unifina/service/ProductService.groovy
kare: CreateProductCommand can contain an empty name. So empty name is replaced with the default.

grails-app/services/com/unifina/service/CanvasService.groovy
kare: 👍Added DEFAULT_NAME

src/groovy/com/unifina/api/CreateProductCommand.groovy
kare: @hpihkala Incoming JSON string can contain en empty name. Should Command objects do just validation or also other tasks?

PR #513, opened 2018-10-04
CORE-1550: Re-add old fix, try eager fetching

jtakalai:
was added in https://github.com/streamr-dev/engine-and-editor/commit/eef4def20807d3a71259d83b343cc34051a99368 but modified in https://github.com/streamr-dev/engine-and-editor/commit/88cef228c4319d5baea4ffbef9fb5a30b27db82c, now re-adding the first fix, just to see if it helps. This is an old problem with lazy loading of IntegrationKey domain object.

### DIFF
--- a/grails-app/domain/com/unifina/domain/security/IntegrationKey.groovy
+++ b/grails-app/domain/com/unifina/domain/security/IntegrationKey.groovy
@@ -20,6 +20,7 @@ class IntegrationKey implements Serializable {
 	static mapping = {
 		id generator: IdGenerator.name // Note: doesn't apply in unit tests
 		json type: 'text'
+		user lazy: false
 		idInService(index: "id_in_service_and_service_idx")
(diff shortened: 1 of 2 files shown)


grails-app/domain/com/unifina/domain/security/IntegrationKey.groovy
```diff
@@ -20,6 +20,7 @@ class IntegrationKey implements Serializable {
 	static mapping = {
 		id generator: IdGenerator.name // Note: doesn't apply in unit tests
 		json type: 'text'
+		user lazy: false
```
jtakalai: this seemed like the only "transitive" domain object contained in IntegrationKey

jtakalai: (which could be lazy-loaded, that is; or eagerly hopefully from now on)

PR #514, opened 2018-10-05
CORE-1552 fix: subscriptionService should only handle ETHEREUM_ID keys

jtakalai:


jtakalai:
Already got review from @harbu

PR #516, opened 2018-10-15
[CORE-1554] Javascript injection vulnerability in old Editor

Aapzu:
- Changed `$(elementAsHtml)` to `$(tagName, attributes)` where necessary (basically where there was any variable injections. Example:
```
// old
var element = $("<div attr1="foo"><a attr2="bar">" + content + "</a></div>")

// new
var a = $('<a/>', {
  attr2: 'bar',
  text: content, // attribute text is escaped by jQuery
})
var element = $('<div/>', {
  attr1: 'foo',
}).append(a)
```
- Change occurences of `$el.html(text)` to `$el.text(text)` (`.text()` escapes the content) where `.html()` cleary was not needed
- Change occurences of `$el.html(jQueryElement)` to `$el.empty().append(jQueryElement)`

I went through the features with changed code but would also appreciate if someone else tried them as well:) (eg. `.append('<div/>', { ... })` instead of `.append($('<div/>', { ... })` ignores the second parameter totally and appends an empty element)

### DIFF
--- a/web-app/js/unifina/signalPath/generic/emptyModule.js
+++ b/web-app/js/unifina/signalPath/generic/emptyModule.js
@@ -42,7 +42,10 @@ SignalPath.EmptyModule = function(data, canvas, prot) {
 	function createDiv() {
 		var buttons = []
 
-		prot.div = $("<div id='"+prot.id+"' class='component module context-menu "+prot.type+"'></div>");
+		prot.div = $("<div/>", {
+			id: prot.id,
+			class: 'component module context-menu ' + prot.type,
+		})
 		prot.div.data("spObject",prot);
 		
 		// Set absolute position
@@ -54,7 +57,10 @@ SignalPath.EmptyModule = function(data, canvas, prot) {
 		
 		// Create title
 		var moduleName = prot.jsonData.displayName || prot.jsonData.name
-		prot.title = $("<span class='modulename'>" + moduleName + "</span>");
+		prot.title = $("<span/>", {
+			class: 'modulename',
+			text: moduleName,
+		})
 		prot.header.append(prot.title);
         
         // If there are options, create options editor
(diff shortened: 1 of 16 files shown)


web-app/js/unifina/signalPath/specific/chartInput.js
```diff
@@ -22,8 +22,10 @@ SignalPath.ChartInput = function(json, parentDiv, module, type, pub) {
     // Use 1-based index for display to the user
     var displayedAxis = json.yAxis + 1
     // Cycle Y-axis button
-    var $yAxisSelectorButton = $("<div class='y-axis-number btn "+btnDefaultClass+" btn-xs "+popoverClass+" popover-colorful'></div>")
-    
+    var $yAxisSelectorButton = $('<div/>', {
```
hpihkala: While manually testing, I noticed that the y-axis selector button is not shown beside the chart input at all! Changes look ok, so not sure if it's related to these changes or some earlier ones?

web-app/js/unifina/streamr-chart/streamr-chart.js
```diff
@@ -70,7 +70,10 @@ function StreamrChart(parent, options) {
 		]
 
 		rangeConfig.forEach(function(c) {
-			var $option =  $("<option value='"+c.range+"'>"+c.name+"</option>")
+			var $option =  $('<option/>', {
```
hpihkala: While manually testing, I noticed that the chart range dropdown initially shows an empty label. Changes look ok, so not sure if it's related to these changes or some earlier ones?

web-app/js/unifina/signalPath/generic/emptyModule.js
```diff
@@ -670,12 +689,14 @@ $('body').contextMenu({
 		menu.data('spObject', target.data('spObject'))
 
 		var prot = menu.data('spObject');
-		var html = prot.getContextMenu(menu.data('target')).map(function(item) {
-			return '<li><a tabindex="-1" data-cmd="' +
-				item.cmd+'" href="#">' +
-				item.title+'</a></li>';
-		}).join('');
-    	menu.html(html)
+		menu.append(prot.getContextMenu(menu.data('target')).map(function(item) {
```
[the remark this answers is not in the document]
Aapzu: Good catch!

web-app/js/unifina/signalPath/specific/chartInput.js
Aapzu: After investigating a bit it seems that this file is never added to the UI (it's not in `ApplicationResources.groovy`)

web-app/js/unifina/streamr-chart/streamr-chart.js
Aapzu: #517 opened

web-app/js/unifina/signalPath/specific/chartInput.js
Aapzu: #518 created

hpihkala:
@Aapzu LGTM! Thanks for looking at the other discovered issues too!

PR #518, opened 2018-10-17
[CORE-1557] Y-axis selector button not visible in chart

Aapzu:
Created a new issue, branch and PR since I wasn't sure what changes needed to be done in order this to work (it has never been used). It worked out-of-the-box so this PR ended up to be only 1 line:)


Aapzu:
@hpihkala in [`chartModule.js`'s `addInput`](https://github.com/streamr-dev/engine-and-editor/blob/master/web-app/js/unifina/signalPath/specific/chartModule.js#L65) we override `prot.addInput` to use `Signalpath.ChartInput`, but in [`genericModule`'s `addInput`](https://github.com/streamr-dev/engine-and-editor/blob/master/web-app/js/unifina/signalPath/generic/genericModule.js#L326) we default the parameter to `SignalPath.Input` if the `clazz` we're trying to use (`Signalpath.ChartInput` in this case) doesn't exist. That's why it still rendered a normal input instead of fully breaking.
Working functional tests would've found that, though 🤷

PR #523, opened 2018-10-23
profile/update endpoint can't update username anymore

hpihkala:
A security researcher discovered a bug in the old `ProfileController#update` action, which allowed the username to be changed. This bug combined with CSRF vulnerability enables an attacker to take over accounts by tricking a user into a malicious site while they are logged in, and then making a request to the faulty endpoint.

- This update only fixes the bug in the endpoint. The app still remains vulnerable to CSRF attacks in general (no other ways to exploit it with critical consequences are currently known).
- The new single-page app, once released, will use token-based sessions instead of cookie-based sessions. That will be the ultimate fix to CSRF. Probably not worth it to ship a full CSRF fix to the old app anymore.

### DIFF
--- a/grails-app/controllers/com/unifina/controller/security/ProfileController.groovy
+++ b/grails-app/controllers/com/unifina/controller/security/ProfileController.groovy
@@ -20,8 +20,7 @@ class ProfileController {
 	static defaultAction = "edit"
 
 	static allowedMethods = [
-		update: "POST",
-		regenerateApiKey: "POST"
+		update: "POST"
 	]
 
 	def edit() {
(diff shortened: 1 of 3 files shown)


grails-app/controllers/com/unifina/controller/security/ProfileController.groovy
```diff
@@ -38,28 +44,6 @@ class ProfileController {
 		}
 	}
 
-	def regenerateApiKey() {
```
hpihkala: unused code removed

grails-app/views/profile/edit.gsp
```diff
@@ -24,23 +24,6 @@
 		</div>
 	</div>
 
-	<r:script>
```
hpihkala: unused code removed

PR #524, opened 2018-10-30
[CORE-1564] - Request Metamask Account Permission (profile settings page)

fonty1:
Requesting permission to read the account information from the Metamask Provider, to be compliant with the upcoming breaking change coming from the Metamask extension. Since this will be washed away in the coming weeks with the new userpages, not much time was spent on the UX.

Affects: Profile Page Settings

Behaviour: 
New Metamask client:
screen shot 2018-10-30 at 20 36 47

screen shot 2018-10-30 at 20 39 37

Old Metamask behaviour:
No change

https://github.com/MetaMask/metamask-extension/pull/4703
https://eips.ethereum.org/EIPS/eip-1102

### DIFF
--- a/web-app/react-app/tests/utils/web3Provider.test.js
+++ b/web-app/react-app/tests/utils/web3Provider.test.js
@@ -53,17 +53,28 @@ describe('web3Provider', () => {
         })
     })
     describe('getWeb3', () => {
+        let oldGlobalWeb3
+        let oldGlobalEthereum
+
         beforeEach(() => {
+            oldGlobalWeb3 = global.web3
+            oldGlobalEthereum = global.ethereum
             global.web3 = undefined
             global.ethereum = undefined
         })
+
+        afterEach(() => {
+            global.web3 = oldGlobalWeb3
+            global.ethereum = oldGlobalEthereum
+        })
+
         it('must return the web3 object without a provider when metamask does not provide it', () => {
             const web3 = getWeb3()
             expect(web3.currentProvider).toEqual(null)
         })
         it('must return the web3 object with the window.web3.currentProvider provider if it is available/defined', () => {
             // 'legacy' metamask web3 injection scenario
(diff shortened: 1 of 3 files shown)


fonty1:
@Aapzu Implementation updated to match the latest changes in the PR. 
Metamask build to test with:
https://github.com/MetaMask/metamask-extension/pull/4703#issuecomment-432346488

Enable Privacy mode in Metamask settings to watch it fly.

fonty1:
bump @Aapzu : )

web-app/react-app/utils/web3Provider.js
```diff
@@ -15,8 +16,13 @@ export class StreamrWeb3 extends Web3 {
     isEnabled = (): boolean => !!this.currentProvider
```
Aapzu: Should we also change the way this function works? Or just let it be / remove it

web-app/react-app/tests/utils/web3Provider.test.js
```diff
@@ -53,8 +53,26 @@ describe('web3Provider', () => {
...
+            global.web3 = undefined
+            global.ethereum = undefined
+        })
+        it('must return the web3 object without a provider when metamask does not provide it', () => {
+            const web3 = getWeb3()
+            expect(web3.currentProvider).toEqual(null)
+        })
+        it('must return the web3 object with the window.web3.currentProvider provider if it is available/defined', () => {
+            // 'legacy' metamask web3 injection scenario
+            global.web3 = Web3
```
Aapzu: Technically `global.web3` shouldn't be `Web3`, but `new Web3()`. Not a big thing, though. (and also using just an object would be enough)

web-app/react-app/utils/web3Provider.js
fonty1: This can be left as is.

PR #526, opened 2018-11-11
Improvements to node IP address autodetection

hpihkala:
When developing offline (e.g. on a plane), the app refuses to start because the IP address detection fails. Also in some other situations (e.g. node behind a NAT) it may be useful to force the node to think that it's on some given IP address. 

Fixed by adding a config parameter `streamr.ip.address.force` which can be set via a system property of the same name.


harbu:
Ran into problems with this IP detection thing again. Couldn't we just whitelist the entire private address range https://en.wikipedia.org/wiki/Private_network#Private_IPv4_addresses ?

hpihkala:
> Ran into problems with this IP detection thing again. Couldn't we just whitelist the entire private address range https://en.wikipedia.org/wiki/Private_network#Private_IPv4_addresses ?

@harbu Can you give details on the problems you're having? Sure we could whitelist all the private address ranges by default.

harbu:
Problem was that docker ip address had changed (for whatever reason). I keep whitelisting private ip addresses to fix this problem every now and then. Just whitelisting everything (or just checking if an ip is private if there exists such a method in Java) would be optimal.

hpihkala:
Rewrote the logic to something much better:

- If an IP address to use is explicitly configured, use that.
- Else find any IPv4 adderss on a non-localhost, non-virtual, non-p2p interface. Use the first one found. If there are multiple candidates, log a warning.
- If no candidates are found, use localhost and log a warning.

PR #533, opened 2018-11-28
Add admin endpoint that returns current Grails config

hpihkala:
Helps debug problems in staging/production


hpihkala:
@harbu Good point. I checked and there is only 1 admin user in prod. Merging this now.

PR #534, opened 2018-11-30
CORE-1591: admin-only user creation API endpoint

harbu:
Admin-only API endpoint for creating users. Handy for creating isolated user environments for well encapsulated integration tests. POST `/api/v1/users` with body
```
username: "something@something.com",
password: "password",
name: "Full Name",
timezone: "UTC" [optional]
```

### DIFF
--- a/grails-app/conf/UrlMappings.groovy
+++ b/grails-app/conf/UrlMappings.groovy
@@ -57,6 +57,7 @@ class UrlMappings {
 		"/api/v1/modules"(resources: "moduleApi")
 		"/api/v1/modules/$id/help"(controller: "moduleApi", action: "help")
 
+		"/api/v1/users"(resources: "userApi")
 		"/api/v1/users/me"(controller: "userApi", action: "getUserInfo")
 		"/api/v1/users/me/keys"(resources: "keyApi", excludes: ["create", "edit", "update"]) { resourceClass = SecUser }
 		"/api/v1/users/me/products"(controller: "productApi", action: "index") { operation = Permission.Operation.SHARE }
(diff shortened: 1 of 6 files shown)


harbu:
Is the Ethereum authentication flow already merged to master? How to use it to create a user? I think we should go with that if username/password-based authentication will be dropped.

hpihkala:
Yeah it's in master. You'd use it with the latest `streamr-client` by doing something like this (@mthambipillai can help):

```javascript
const privateKey = ... // Random hex string of 256 bytes (64 hex characters)
const client = new StreamrClient({
    auth: {
        privateKey,
    }
})

// Then you can simply make API calls using methods on the client
client.createStream(...)

// or if you just need a valid session token to make API calls yourself, you can do:
client.session.getSessionToken() // returns a Promise that resolves with session token

// Then in your API calls set the header:
// Authorization: Bearer your-session-token-here
```

Any method of generating the private key should be fine as long as it's correct length, but you can also use the [web3 library](https://web3js.readthedocs.io/en/1.0/web3-eth-accounts.html) if you want to be sure.

hpihkala:
@timoxley can you test if the above approach of key-based user auto-generation works for you? If you need to generate a new private key for each test, you can use e.g. the [web3](https://web3js.readthedocs.io/en/1.0/web3-eth-accounts.html) library, or any [other method](https://stackoverflow.com/a/27747377/1073193) of generating a random hex string 64 characters long.

harbu:
Apparently in the rest-e2e-tests I have used the following snippet to generate private keys
```
const Web3 = require('web3')
const keythereum = require('keythereum')

...

const generatedKey = keythereum.create()
const privateKey = '0x' + generatedKey.privateKey.toString('hex')
const publicAddress = '0x' + ethereumJsUtil.privateToAddress(generatedKey.privateKey).toString('hex')
```

Dunno if that is of any help.

harbu:
I'm closing this for now as it seems that the same effect can be achieved by the Ethereum-based authentication flow.

PR #539, opened 2018-12-19
Fix message parsing regression in master

hpihkala:
**Why it broke**

After some binary protocol refactoring in #528, `StreamrBinaryMessageWithKafkaMetadata` is no longer a subclass of `StreamrBinaryMessage`, but instead the former contains the latter.

`StreamrBinaryMessageParser` expects `StreamrBinaryMessage`s as input - a condition previously fulfilled by both `StreamrBinaryMessageWithKafkaMetadata` and `StreamrBinaryMessage` due to inheritance, but after the refactor the class would not accept `StreamrBinaryMessageWithKafkaMetadata` anymore, failing only at runtime due to some generics at play.

**Why it wasn't caught by tests**

Integration test `RunCanvasSpec` tests that the overall message pipeline is working, but for some reason it was `@Ignore`d in [this commit](https://github.com/streamr-dev/engine-and-editor/commit/34a11beb4ef3de7c1a6e9e414eff807a3abb72b2), meaning that it is not being run.

**What's the fix**

- Refactored `RedisMessageSource` and related components to output `StreamrBinaryMessage`s instead of `StreamrBinaryMessageWithKafkaMetadata`. 
- Re-enabled `RunCanvasSpec` in a separate PR #540, pending fixes to CI issues.

### DIFF
--- a/src/java/com/unifina/feed/redis/MultipleRedisMessageSource.java
+++ b/src/java/com/unifina/feed/redis/MultipleRedisMessageSource.java
@@ -1,5 +1,6 @@
 package com.unifina.feed.redis;
 
+import com.unifina.data.StreamrBinaryMessage;
 import com.unifina.domain.data.Feed;
 import com.unifina.feed.AbstractMessageSource;
 import com.unifina.feed.MessageRecipient;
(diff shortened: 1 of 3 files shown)


hpihkala:
Should maybe re-disable RunCanvasSpec. @harbu disabled it in the spring due to CI problems, and the problems seem to still persist.

PR #540, opened 2018-12-20
WIP: Fix RunCanvasSpec

hpihkala:
Bring back `RunCanvasSpec`, which was originally `@Ignore`d by @harbu months ago due to troubles on Travis.

Two bugs were found and fixed:
- `RunCanvasSpec` was transactional and ended up blocking database transactions executed by the system under test
- Kafka didn't work due to the equivalent of `streamr-docker-dev bind-ip` missing from Travis config

### DIFF
--- a/grails-app/services/com/unifina/service/NodeService.groovy
+++ b/grails-app/services/com/unifina/service/NodeService.groovy
@@ -34,7 +34,7 @@ class NodeService {
 			if (addresses.size() == 1) {
 				cachedIp = addresses[0].hostAddress
 			} else if (addresses.size() > 1) {
-				log.warn("Multiple IPs found: ${addresses*.hostAddress}. To select which address to use, please set the streamr.node.ip system property.")
+				log.warn("Multiple IPs found: ${addresses*.hostAddress}. By default the first one will be used. To select another aaddress, please set the streamr.node.ip system property.")
 				cachedIp = addresses[0].hostAddress
 			} else {
 				// Fallback to localhost
@@ -43,6 +43,7 @@ class NodeService {
 			}
 		}
 
+		log.info("Using IP address: " + cachedIp)
 		return cachedIp
 	}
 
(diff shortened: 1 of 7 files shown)


harbu:
It seems to be broken in local dev environment too. If I start up a new Docker environment from scratch and then proceed to run this test, on the first run it will fail. Subsequent runs pass.

hpihkala:
Compared logs between healthy and unhealthy runs.

Unhealthy:

- Runner thread gets stuck (no further logging for 30 seconds after below line)
`2018-12-20 12:58:50,474 [lettuce-nioEventLoop-4-1] INFO  redis.RedisMessageSource  - Subscribed run-canvas-spec-0 on host 127.0.0.1`
- `SignalPathRunner.waitRunning()` times out after 30 seconds
- Canvas start gets aborted due to timeout

**Clearly the runner thread blocks waiting for something, but I haven't yet found out what.**

Healthy:

- Message that the realtime queue starts right after the subscribe message

```
2018-12-20 12:41:10,327 [lettuce-nioEventLoop-4-1] INFO  redis.RedisMessageSource  - Subscribed run-canvas-spec-0 on host 127.0.0.1
2018-12-20 12:41:10,395 [SignalPathRunner [run-canvas-spec]] INFO  data.RealtimeEventQueue  - The realtime event queue is starting!
```

hpihkala:
@harbu Do you also get those Kafka errors when the test fails locally? (Would be good to establish whether the two symptoms are related)

harbu:
I was only looking at E&E logs but I didn't see any Kafka-related errors? Are they supposed to pop in E&E, Kafka, or ZooKeeper logs?

hpihkala:
Good news: 
- Managed to fix the Kafka errors by adding the `bind-ip` equivalent to the travis script. 
- The test now terminates instead of running until Travis kills it.
Bad news:
- The test still fails with the same symptoms as before.

hpihkala:
The issue was caused by `RunCanvasSpec` being transactional by default. The transaction that wraps the test was blocking other transactions executed in the `SignalPathRunner` thread. Fixed by setting the spec to `transactional = false`.

PR #541, opened 2019-01-06
Cleanup and update API docs

hpihkala:


hpihkala:
@githubwei Thanks for your comments!

> Should we add link to Reddit, Telegram and/or generic email contact@streamr.com?

Yes! Added!

> In the Authentication section, it seems link to profile page was incorrect

It's OK, the markdown gets processed before rendering. The tag you see defines a relative target, which gets replaced with the correct absolute URL (which depends on where we are running: production, staging, locally...).

> In the API explorer section are we planning to add the AOX link there soon?

The current API explorer based on Swagger UI gets appended to this content, so no trouble there! (This file belongs to larger context, it's not meant to be a self-sufficient document.)

We'll replace the current API explorer with a link to the OAX variant when/if it's ready for deployment. We found out that the OAX tool is abandoned and is not being developed further, which is a shame. Matthew is looking into whether we can still make it work for us.

PR #542, opened 2019-01-07
Fix anonymous key login + status codes (400 vs 401)

mthambipillai:


grails-app/domain/com/unifina/domain/security/Key.groovy
```diff
@@ -53,4 +53,9 @@ class Key implements Userish {
 	Userish resolveToUserish() {
 		return user ?: this
 	}
+
+	// Cannot use @Override because of bug in Grails 2.3 and later
```
hpihkala: Is there any link to the bug description/discussion?

grails-app/services/com/unifina/service/SessionService.groovy
```diff
@@ -11,19 +13,29 @@ class SessionService {
 
 	KeyValueStoreService keyValueStoreService
 
-	SessionToken generateToken(SecUser user) {
-		SessionToken sk = new SessionToken(TOKEN_LENGTH, user, TTL_HOURS)
-		keyValueStoreService.setWithExpiration(sk.getToken(), sk.getUser().id.toString(), TTL_HOURS * 3600)
+	SessionToken generateToken(Userish userish) {
+		SessionToken sk = new SessionToken(TOKEN_LENGTH, userish, TTL_HOURS)
+		keyValueStoreService.setWithExpiration(sk.getToken(), userish.getClassAndId(), TTL_HOURS * 3600)
```
hpihkala: IMO it would actually be nicer to drop the `getClassAndId()` method from the `Userish` interface and implement the logic in `SessionService` instead. 2 reasons:

- The inverse logic (`String` to `Userish`) is here, so for symmetry `Userish` to `String` should be too.
- `SessionService` is the only user of that logic.

For readability, the conversion in both directions could be wrapped in symmetric private methods if you like:
- `String userishToString(Userish)`
- `Userish stringToUserish(String)`

PR #543, opened 2019-01-08
(wip) Refactor signup, register, and forgot password endpoints under /api/v1

hpihkala:
Endpoints related to new user registration and password recovery were moved under /api/v1 to prepare for the new frontend app. 

**Mostly** this just changes the URLs, but there is one breaking change in functionality: completing registration doesn't automatically sign the user in anymore (doesn't set the spring security session cookie). 

Instead, the frontend should: 
- After the call to `POST /api/v1/users` succeeds, **explicitly start the user's session by calling the login endpoint** with the username and password (`/j_spring_security_check` in this old frontend, `/api/v1/login/password` in the new frontend).
- Then redirect the user to the default view (`/canvas/editor` in this old frontend). 

@mondoreale this frontend functionality is still missing from this PR. Could you help implement it?


mondoreale:
@hpihkala, I can take it from here, thanks!

mondoreale:
It's going to be an interesting experience to deploy this little guy. https://github.com/streamr-dev/streamr-platform/pull/199 uses the old non-api endpoints for signup/passwords cause this is what's on staging e&e atm. And we want to test it on staging cause local dockers don't send emails.

An example of how the deployment could look like:
1. We make a staging release of e&e w/ the new endpoints.
2. We update endpoints in the platform and also release it to staging.
3. We test if apps play well against each other.
4. We deploy both at pretty much the same time to production.

Clearly a coordination challenge, but we should be fine!

PR #545, opened 2019-01-10
PackageAccessHelper: prevent System etc. in dev env

harbu:


jtakalai:
How are these prevented in production?

hpihkala:
Fixed in a more general way by reintroducing `SecurityManager` and `Policy` in #546. Closing this.

PR #549, opened 2019-01-11
Fix add staging env

mikhaelsantos:


Dockerfile.stg
```diff
@@ -0,0 +1,26 @@
...
+# Install dependencies
+#   bash: required by wait_for_it.sh script
+#   mysql: required for waiting for database dumps to be done
+RUN apk update
+RUN apk add bash mysql-client
+RUN sed -i "s/port=\"8080\"/port=\"8081\"/g" /usr/local/tomcat/conf/server.xml
+# Copy wait-for-it.sh script
+COPY wait-for-it.sh /usr/local/tomcat/bin/wait-for-it.sh
+
+# Copy pre-built unifina-core.war into container as unifina-core.war
```
hpihkala: unifina-core.war -> streamr-core.war

Dockerfile.stg
```diff
@@ -0,0 +1,26 @@
...
+COPY target/ROOT.war /usr/local/tomcat/webapps/streamr-core.war
+
+ENV CATALINA_OPTS -Dstreamr.database.user=root \
+    -Dstreamr.database.password=password \
+    -Dstreamr.database.host=mysql \
+    -Dstreamr.kafka.bootstrap.servers=kafka:9092 \
+    -Dstreamr.ui.server=ws://127.0.0.1:8890/api/v1/ws \
```
hpihkala: This and the below line should point to urls that reach the data-api of the deployment. If this is for staging, should it be some non-localhost address?

PR #558, opened 2019-01-24
CORE-1566: Admin endpoint to detect stale products

kare:


grails-app/controllers/com/unifina/controller/api/ProductApiController.groovy
```diff
@@ -31,6 +36,24 @@ class ProductApiController {
...
 	ProductImageService productImageService
 
+	@GrailsCompileStatic
+	@StreamrApi(allowRoles = AllowRole.ADMIN)
+	def staleProducts() {
+		List<Product> products = productService.list(new ProductListParams(publicAccess: true), loggedInUser())
+		int days = params.int("days", 2)
+		Date threshold = DateUtils.addDays(new Date(), -days)
+		List<ProductService.StaleProduct> results = productService.findStaleProducts(products, threshold)
+		System.out.println(results)
```
hpihkala:
```suggestion
```

grails-app/controllers/com/unifina/controller/api/ProductApiController.groovy
```diff
@@ -31,6 +36,24 @@ class ProductApiController {
...
+		def json = results.collect { stale ->
+			[id: stale.product.id, name: stale.product.name, streams: stale.streams.collect { stream ->
+				StreamrMessage msg = stream.findMostRecentMessage()
```
hpihkala: Aren't you already fetching the most recent message in `findStaleProducts`? Why not store it in the results to avoid having to look it up again?

grails-app/services/com/unifina/service/CassandraService.groovy
```diff
@@ -120,6 +121,29 @@ class CassandraService implements DisposableBean {
...
+			final StreamrBinaryMessage bin = meta.getStreamrBinaryMessage()
+			final StreamrMessage msg = bin.toStreamrMessage()
+			messages.add(msg)
+		}
+		if (messages.size() < 1) {
+			return null
+		}
+		Date now = new Date(0)
+		StreamrMessage latest = null
+		for (int i = 0; i < messages.size(); i++) {
```
hpihkala:
```suggestion
		for (StreamrMessage m : messages) {
```

grails-app/services/com/unifina/service/CassandraService.groovy
```diff
@@ -120,6 +121,29 @@ class CassandraService implements DisposableBean {
...
+			StreamrMessage m = messages.get(i)
+			if (m.getTimestamp().after(now)) {
```
hpihkala: **(untested)** Could be simplified to (and variable `now` deleted):
```
			if (latest == null || m.getTimestamp().after(latest)) {
				latest = m
			}
```

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -14,8 +15,82 @@ class ProductService {
 	ApiService apiService
 	PermissionService permissionService
 	SubscriptionService subscriptionService
+	CassandraService cassandraService
 	Random random = ThreadLocalRandom.current()
 
+	static class StreamWithMessages {
```
hpihkala: Why do we need to collect a list of messages? Looks like (line 70) that the list will always contain exactly 1 message anyway. Maybe just change the class to `StreamWithLatestMessage`:

```
static class StreamWithLatestMessage {
    Stream stream
    StreamrMessage latestMessage

    String toString() {
        // ...
    }
}
```
Also see comments below

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -14,8 +15,82 @@ class ProductService {
...
+	static class StaleProduct {
+		Product product
+		final List<StreamWithMessages> streams = new ArrayList<>()
+		String toString() {
+			return String.format("StaleProduct[product=%s, streams=%s]", product, streams.toString())
+		}
+	}
+
+	// ???This will skip duplicate product in the results in case of product with two streams???
+	boolean contains(List<StaleProduct> products, StaleProduct stale) {
```
hpihkala: I think this can be removed, see suggestion below

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -14,8 +15,82 @@ class ProductService {
...
+			StaleProduct stale = new StaleProduct()
+			stale.product = p
+			// Get messages from streams
+			for (Stream s : p.getStreams()) {
+				StreamrMessage msg = cassandraService.getLatestFromAllPartitions(s)
+				if (msg != null) {
+					StreamWithMessages sm = new StreamWithMessages()
+					sm.stream = s
+					sm.messages.add(msg)
+					stale.streams.add(sm)
```
hpihkala: I suggest you filter while iterating to simplify this considerably:

- Only create and add a `StreamWithMessages` to the `StaleProduct stale` if the stream is stale
- A stream if stale if the newest message is old enough + note that streams with no data should be considered stale too
- Only add `stale` to `staleProducts` if `stale.streams.size() > 0`


Something like (untested):
```
// ...
for (Stream s : p.getStreams()) {
    StreamrMessage msg = cassandraService.getLatestFromAllPartitions(s)
    if (msg != null && msg.getTimestamp().before(threshold)) { // logic corrected: after -> before!
        StreamWithLatestMessage sm = new StreamWithLatestMessage(stream: s, latestMessage:msg)
        stale.streams.add(sm)
    } else if (msg == null) {
        // Streams with no messages are considered stale, with latestMessage null
        StreamWithLatestMessage sm = new StreamWithLatestMessage(stream: s)
        stale.streams.add(sm)
    }
}

if (!stale.streams.isEmpty()) {
    staleProducts.add(stale)
}

// Remove the separate filtering phase
```

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -14,8 +15,82 @@ class ProductService {
...
+			staleProducts.add(stale)
+		}
+
+		// Find products whose streams haven't received new messages within threshold days
+		final List<StaleProduct> results = new ArrayList<>()
+		for (StaleProduct sp : staleProducts) {
+			for (StreamWithMessages sm : sp.streams) {
+				StreamrMessage m = sm.findMostRecentMessage()
+				if (m != null) {
+					if (m.getTimestamp().after(threshold)) {
```
hpihkala: Logic error here? The product if stale if the _latest message_ is `before` (older than) the threshold point in time, not `after`?! If I'm right there must be something wrong with the unit test too.

Already fixed in the cleanup suggestion above.

test/unit/com/unifina/service/ProductServiceSpec.groovy
```diff
@@ -61,6 +64,83 @@ class ProductServiceSpec extends Specification {
...
+		1 * service.cassandraService.getLatestFromAllPartitions(s2) >> m2
+		1 * service.cassandraService.getLatestFromAllPartitions(s3) >> m3
+		1 * service.cassandraService.getLatestFromAllPartitions(s4) >> m4
+		1 * service.cassandraService.getLatestFromAllPartitions(s5) >> m5
+		1 * service.cassandraService.getLatestFromAllPartitions(s6) >> m6b
+		1 * service.cassandraService.getLatestFromAllPartitions(s7) >> m7
+		1 * service.cassandraService.getLatestFromAllPartitions(s8) >> m8
+		1 * service.cassandraService.getLatestFromAllPartitions(s9) >> m9
+
+		results.size() == 5
```
hpihkala: Logic is exactly the opposite: should return the products with no new messages in 2 days (or no messages at all).

Should return the products with the following streams:
```
		Stream s3 = newStream("s3", "Storm Stream") // This stream has a message three days ago
		Stream s6 = newStream("s6", "Mainframe Stream") // This stream has a message two weeks ago
		Stream s6b = newStream("s6b", "Mainframe B Stream") // This stream has a message one week ago
		Stream s7 = newStream("s7", "Barometer Stream") // This stream doesn't have any messages
```

test/unit/com/unifina/service/ProductServiceSpec.groovy
```diff
@@ -61,6 +64,83 @@ class ProductServiceSpec extends Specification {
...
+		results.find { it.product.id == "a" && it.product.name == "Air quality" && it.streams.size() == 1 /*&& it.streams.get(0).id == "s1"*/ }
+		results.find { it.product.id == "b" && it.product.name == "Wind speed" && it.streams.size() == 1 /*&& it.streams.get(0).id == "s2"*/ }
+		results.find { it.product.id == "d" && it.product.name == "Hacked computers" && it.streams.size() == 1 /*&& it.streams.get(0).id == "s4"*/ }
+		results.find { it.product.id == "e" && it.product.name == "Time machine" && it.streams.size() == 1 /*&& it.streams.get(0).id == "s5"*/ }
+		results.find { it.product.id == "h" && it.product.name == "Product with two streams" && it.streams.size() == 2 /*&& it.streams.get(0).id == "s8" && it.streams.get(1).id == "s9"*/ }
+	}
```
hpihkala: Please also include a test product with 2 streams: 1 stale (latest message e.g. 2 weeks ago) and 1 non-stale (latest message a few hours ago).

The product should be returned by `service.findStaleProducts` but only the stale stream should be listed.

PR #559, opened 2019-01-24
Message id version 30

mthambipillai:


grails-app/controllers/com/unifina/controller/data/StreamController.groovy
```diff
@@ -165,7 +165,7 @@ class StreamController {
 					response.status = 500
 					render([success: false, redirect: createLink(action: 'confirm', params: [id: params.id, file: temp.getCanonicalPath()])] as JSON)
 				} else {
-					Map updatedConfig = streamService.importCsv(csv, stream)
+					Map updatedConfig = streamService.importCsv(csv, stream, user.username)
```
hpihkala: This might need some more discussion/thought, but my first instinct is that we should probably use the `user.id` as the id for non-Ethereum users to avoid exposing the user's email address to subscribers.

The logic for what to use as the `producerId` could be abstracted by `SecUser`, as now you notice that you'll have to change quite many places to change the logic. Something like:

```
String getProducerId() {
    // If Ethereum user return username (=address)
    // Else return id.toString()
}
```

grails-app/controllers/com/unifina/controller/data/StreamController.groovy
```diff
@@ -207,7 +207,7 @@ class StreamController {
 			def format = params.customFormat ?: params.format
 			try {
 				CSVImporter csv = new CSVImporter(file, fields, index, format)
-				Map config = streamService.importCsv(csv, stream)
+				Map config = streamService.importCsv(csv, stream, user.username)
```
hpihkala: Here too

grails-app/services/com/unifina/service/CsvUploadService.groovy
```diff
@@ -56,7 +56,7 @@ class CsvUploadService {
 
 		try {
 			CSVImporter csv = new CSVImporter(file, fields, instructions.timestampColumnIndex, instructions.dateFormat)
-			Map newStreamConfig = streamService.importCsv(csv, stream)
+			Map newStreamConfig = streamService.importCsv(csv, stream, user.username)
```
hpihkala: Here too

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -128,40 +120,41 @@ class StreamService {
 
 	// Ref to Kafka will be abstracted out when Feed abstraction is reworked
 
-	void sendMessage(Stream stream, String partitionKey, long timestamp, byte[] content, byte contentType, int ttl=0,
-					 SignatureType signatureType = SignatureType.SIGNATURE_TYPE_NONE, String address = null, String signature = null) {
+	void sendMessage(Stream stream, String partitionKey, long timestamp, String content, StreamMessage.ContentType contentType,
```
hpihkala: Since the sender is now stateful (must maintain `sequenceNumber` and `previousMsgRef`), meaning the sender has a lot of responsibility for constructing a correct message anyway, wouldn't it be nice to just have `sendMessage(StreamMessage)`? Could that be pulled off?

Now you're hard-coding the sequenceNumber to 0 and previousMsgRef to null, and defaulting address to `""`, but would be good to get proper values in them. Passing the responsibility to whoever calls this and exposing just `sendMessage(StreamMessage)` would enable/encourage/force that.

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -128,40 +120,41 @@ class StreamService {
...
-	void sendMessage(Stream stream, @Nullable String partitionKey, Map message, int ttl=0) {
+	void sendMessage(Stream stream, @Nullable String partitionKey, Map message) {
 		String str = gson.toJson(message)
-		sendMessage(stream, partitionKey, System.currentTimeMillis(), str.getBytes(utf8), StreamrBinaryMessage.CONTENT_TYPE_JSON, ttl);
+		sendMessage(stream, partitionKey, System.currentTimeMillis(), str, StreamMessage.ContentType.CONTENT_TYPE_JSON)
 	}
 
-	void saveMessage(Stream stream, String partitionKey, long timestamp, Map message, int ttl, long messageNumber, Long previousMessageNumber) {
-		int streamPartition = partitioner.partition(stream, partitionKey)
+	StreamMessage saveMessage(Stream stream, int streamPartition, long timestamp, long sequenceNumber, String publisherId, Map message, long previousTimestamp, long previousSequenceNumber) {
```
hpihkala: If we expose just `sendMessage(StreamMessage)`, then here too `saveMessage(StreamMessage)` becomes the new standard!

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -189,14 +182,9 @@ class StreamService {
...
-	public Map importCsv(CSVImporter csv, Stream stream) {
-		/**
-		 * Batch-imported rows have negative Kafka offsets to avoid collisions with messages actually produced
-		 * to the stream via Kafka.
-		 */
-		List<Long> latestOffsetByPartition = (0..stream.getPartitions()-1).collect { Integer partition ->
-			StreamrBinaryMessageWithKafkaMetadata msg = cassandraService.getLatestBeforeOffset(stream, partition, 0)
-			return msg ? msg.offset : MINIMUM_SAFE_INTEGER_IN_JAVASCRIPT
+	public Map importCsv(CSVImporter csv, Stream stream, String publisherId) {
+		List<StreamMessage> latestMessageByPartition = (0..stream.getPartitions()-1).collect { Integer partition ->
```
hpihkala: In this method, thanks to the new message ids, I think there's basically no longer a need to know what messages pre-exist or maintain the `latestMessageByPartition` lookup. You can simply have a `long sequenceNumber = 0` and then `sequenceNumber++` after every `saveMessage`.

src/java/com/unifina/feed/AbstractFeedProxy.java
```diff
@@ -75,53 +67,35 @@ public void receive(Message parsedMsg) {
...
-			expected++;
-			FeedEvent<MessageClass, EventRecipientClass>[] events = process(msg);
-
-			if (firstRealQueue == null && checkAge) {
-				firstRealQueue = counter;
-			}
 
-			if (events != null) {
-				// Warn about old events
-				if (checkAge && events.length > 0) {
```
hpihkala: The `checkAge` logic and logged warnings are important and shouldn't be removed! It's currently the only way to see from the logs if there's delay/backpressure in the message queue.

Sure it could be more sophisticated, but...

src/java/com/unifina/feed/StreamrMessage.java
```diff
@@ -3,34 +3,17 @@
 import java.util.Date;
 import java.util.Map;
 
-import com.unifina.data.StreamrBinaryMessageV29.SignatureType;
 import com.unifina.feed.map.MapMessage;
 
 public class StreamrMessage extends MapMessage {
```
hpihkala: I was kind of hoping that it might be possible to kill this class (and `MapMessage` and maybe even `ITimestamped`) and replace them with just `StreamMessage`. I'm not sure what kind of corner case problems you'll run into (maybe something with the time events). I recommend to leave this to be the last comment you address, and see if it is feasible.

grails-app/services/com/unifina/service/StreamService.groovy
mthambipillai: This would work yes. But wasn't the initial logic supposed to order by timestamp and only use the sequence number for conflicting timestamps?

src/java/com/unifina/feed/StreamrMessage.java
mthambipillai: I tried but only managed to use only `StreamMessage` for field detection. Basically a lot of classes require `MessageClass` to implement `ITimestamped`. If `StreamMessage` was the only class used we could easily replace the generics with `MessageClass extends StreamMessage` but there is at least one other class used as `MessageClass`: `RuntimeRequest`.

grails-app/services/com/unifina/service/StreamService.groovy
hpihkala: > But wasn't the initial logic supposed to order by timestamp and only use the sequence number for conflicting timestamps?

Sure, the minimum requirement is that sequence numbers are unique within a timestamp. However fulfilling the stronger property of being unique within a session / csv file / universally does fulfill the minimum requirement plus additionally provide more information which may be useful for human debugging, estimating number of messages in a gap, or similar. Should we establish a best practice that exceeds the minimum requirement, or is it just confusing?

> Also how would we compute the previous timestamp and sequence number for each message?

It's not strictly necessary, it may be better to intentionally break the ref chain between bulk uploads, because they might happen in whatever order. Consider for example this scenario:

- There's existing data up to timestamp t1
- User uploads a batch of data from t3 to t4. First message in batch (t3) would now point to latest existing message (with timestamp t1)
- User uploads another batch of data from t2-t3. First message in batch would point to the same pre-existing message t1. Now both t3 and t2 would have a previousRef of t1.

Historical data uploads present other pretty annoying problems too, for example if the user uploads overlapping CSV files. Let's say file A has data from timestamp t1 to t3 and file B has data from t2 to t4. This leaves us with two parallel message chains, which would be an invalid situation. We could forbid this situation by checking that the timestamp range of the file contains no data in the stream.

src/java/com/unifina/feed/StreamrMessage.java
hpihkala: Yeah, I see! I can think of two quick solutions to make it work with the event queue (which assumes `ITimestamped`s) and still avoid duplicating things with `StreamrMessage`

- Wrap the `StreamMessage` in some class which implements `ITimestamped`:

```
public class WrappedStreamMessage implements ITimestamped {
    private final StreamMessage streamMessage
    public WrappedStreamMessage(StreamMessage streamMessage) {
        this.streamMessage = streamMessage;
    }
    // ...
}
```

- Or move the simple `ITimestamped` interface to the Java client and have `StreamMessage` implement it from the start!

I might go with the 2nd approach because it would allow us to use `StreamMessage`s in the event queue directly, and the `ITimestamped` interface obviously doesn't move any functionality to a wrong place.

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -120,41 +120,43 @@ class StreamService {
...
-			0, address, (Long) null, 0, contentType, content, signatureType, signature)
-
-		String kafkaPartitionKey = "${stream.id}-$streamPartition"
+	void sendMessage(StreamMessage msg) {
+		String kafkaPartitionKey = "${msg.getStreamId()}-${msg.getStreamPartition()}"
 		kafkaService.sendMessage(msg, kafkaPartitionKey)
 	}
 
 	@CompileStatic
 	void sendMessage(Stream stream, Map message) {
```
hpihkala: I meant that we could keep only `sendMessage(StreamMessage msg)` above and delete all the other variants of `sendMessage`. Is that doable? That way we could roll the responsibility for the message chaining and sequence numbering to the caller, which is where it should be.

mthambipillai: I guess it is doable but it will involve a lot a code duplication in many files to basically do the same thing as these functions do. The `gson` object has to be duplicated in every file to convert to JSON strings. I tried and ran into some issue because the `com.google.Gson` class is not serializable and apparently the modules need to be serializable by `ModuleTestHelper.Builder`.

src/java/com/unifina/feed/StreamrBinaryMessageKeyProvider.java
```diff
@@ -1,15 +1,16 @@
...
+import com.streamr.client.protocol.message_layer.StreamMessage;
 import com.unifina.data.IStreamRequirement;
 import com.unifina.domain.data.Feed;
 import com.unifina.utils.Globals;
 
 import java.util.ArrayList;
 import java.util.List;
 
-public class StreamrMessageKeyProvider extends AbstractKeyProvider<IStreamRequirement, StreamrMessage, String> {
+public class StreamrBinaryMessageKeyProvider extends AbstractKeyProvider<IStreamRequirement, StreamMessage, String> {
```
mthambipillai: The name `StreamrBinaryMessageKeyProvider` is not accurate anymore, but it is part of the database migration `2016-09-29-new-data-pipeline.groovy` and renaming it introduced errors. I don't know if it's worth it to do a migration for that or if it's fine to leave this name as is.

hpihkala: Let's just leave it for now. We should remove the whole unnecessary configurability layer at some point, and this will be deleted or revisited at the time.

src/java/com/unifina/signalpath/ModuleWithUI.java
```diff
@@ -68,8 +72,17 @@ private StreamService getStreamService() {
 		return streamService;
 	}
 
-	public void pushToUiChannel(Map msg) {
-		getStreamService().sendMessage(getUiChannel().getStream(), msg);
+	public void pushToUiChannel(Map content) {
+		Stream stream = getUiChannel().getStream();
+		int streamPartition = StreamPartitioner.partition(stream, null);
+		try {
+			StreamMessage msg = new StreamMessageV30(stream.getId(), streamPartition, System.currentTimeMillis(), 0L,
```
hpihkala: Should fill in correct values for `publishedId` and previous message ref

src/java/com/unifina/signalpath/custom/AbstractJavaCodeWrapper.java
```diff
@@ -57,7 +58,11 @@ public void connectionsReady() {
 
 	@Override
 	public void sendOutput() {
-		instance.sendOutput();
+		try {
```
hpihkala: Unnecessary change now?

src/java/com/unifina/signalpath/kafka/SendToStream.java
```diff
@@ -71,19 +74,27 @@ protected boolean allowSideEffectsInHistoricalMode() {
...
 	@Override
-	public void activateWithSideEffects() {
+	public void activateWithSideEffects() throws IOException {
 		ensureServices();
 		Stream stream = streamParameter.getValue();
 		authenticateStream(stream);
-		streamService.sendMessage(stream, inputValuesToMap());
+		int streamPartition = StreamPartitioner.partition(stream, null);
+		StreamMessage msg = new StreamMessageV30(stream.getId(), streamPartition, System.currentTimeMillis(),
+				0L, "", null, 0L,
```
hpihkala: Use correct values for publisher id and previous message ref

src/java/com/unifina/signalpath/kafka/SendToStream.java
```diff
@@ -71,19 +74,27 @@ protected boolean allowSideEffectsInHistoricalMode() {
...
 
 	@Override
-	protected void activateWithoutSideEffects() {
+	protected void activateWithoutSideEffects() throws IOException {
 		Globals globals = getGlobals();
 
 		// Create the message locally and route it to the stream locally, without actually producing to the stream
-		StreamrMessage msg = new StreamrMessage(streamParameter.getValue().getId(), 0, globals.time, inputValuesToMap()); // TODO: fix hard-coded partition
+		int streamPartition = StreamPartitioner.partition(streamParameter.getValue(), null);
+		StreamMessage msg = new StreamMessageV30(streamParameter.getValue().getId(), streamPartition, globals.time.getTime(), 0, "", null, 0L,
```
hpihkala: Use correct values for publisher id and previous message ref

src/java/com/unifina/utils/testutils/FakeStreamService.java
```diff
@@ -21,4 +22,13 @@ public void sendMessage(Stream stream, Map message) {
 		sentMessagesByChannel.get(c).add(message);
 	}
```
hpihkala: `sendMessage(Stream stream, Map message)` doesn't exist anymore in `StreamService`, safe to remove? (It should actually fail compilation due to `@Override` annotation, but it's a Groovy bug that it doesn't fail)

src/java/com/unifina/signalpath/ModuleWithUI.java
```diff
@@ -72,17 +101,30 @@ private StreamService getStreamService() {
 		return streamService;
 	}
 
-	public void pushToUiChannel(Map content) {
-		Stream stream = getUiChannel().getStream();
+	private StreamMessage getMsg(Stream stream, Map content){
 		int streamPartition = StreamPartitioner.partition(stream, null);
+		long timestamp = System.currentTimeMillis();
```
hpihkala: Never use system time in modules, can't work in historical playbacks. Always use event time available in `getGlobals().time`

src/java/com/unifina/signalpath/ModuleWithUI.java
```diff
@@ -72,17 +101,30 @@ private StreamService getStreamService() {
...
+		long sequenceNumber = getNextSequenceNumber(stream.getId(), timestamp);
+		SecUser user = SecUser.getViaJava(getGlobals().getUserId());
```
hpihkala: Can't do a database lookup for every event, too slow! The user can't change during runtime, so you can cache it into a variable and reuse it. For safe serialization/deserialization, you might wanna make it `transient` and re-fetch it if it's null:

```
private transient SecUser cachedUser = null;

private SecUser getUser() {
    if (cachedUser == null) {
        cachedUser = SecUser.getViaJava(getGlobals().getUserId());
    }
    return cachedUser;
}

    // ... then in getMsg
    SecUser user = getUser();
```

src/java/com/unifina/signalpath/ModuleWithUI.java
```diff
@@ -32,13 +34,38 @@
...
 	private static final Logger log = Logger.getLogger(ModuleWithUI.class);
 
+	private Map<String,Long> previousTimestamps = new HashMap<>();
+	private Map<String,Long> previousSequenceNumbers = new HashMap<>();
+
 	public ModuleWithUI() {
 		super();
 	}
 
+	private long getNextSequenceNumber(String streamId, long timestamp) {
```
hpihkala: Can't this functionality (of keeping track of latest MessageRef and sequence number by stream) be abstracted into some utility class
`public class MessageChainUtil implements Serializable { ... }`
and reused here and in `SendToStream`?

Also the chaining should happen by stream-partition I guess?

```
MessageChainUtil msgChainUtil = new MessageChainUtil();

// ...

// Returns the latest known MessageRef (created with .nextMessageRef(...))
MessageRef previousRef = msgChainUtil.getLatestMessageRef(stream, partition)

// Creates and caches a new MessageRef for stream-partition, increments sequence number
MessageRef thisMsgRef = msgChainUtil.nextMessageRef(stream, partition, timestamp)
```

(Also please add unit tests for this util class)

src/java/com/unifina/signalpath/ModuleWithUI.java
```diff
@@ -72,17 +101,30 @@ private StreamService getStreamService() {
...
 	}
 
-	public void pushToUiChannel(Map content) {
-		Stream stream = getUiChannel().getStream();
+	private StreamMessage getMsg(Stream stream, Map content){
 		int streamPartition = StreamPartitioner.partition(stream, null);
+		long timestamp = System.currentTimeMillis();
+		long sequenceNumber = getNextSequenceNumber(stream.getId(), timestamp);
+		SecUser user = SecUser.getViaJava(getGlobals().getUserId());
+		String publisherId = user == null ? "" : user.getPublisherId();
```
hpihkala: AFAIK user should never be null. Canvases always run as some user.

src/java/com/unifina/signalpath/kafka/SendToStream.java
```diff
@@ -43,13 +43,40 @@
 	private boolean sendOnlyNewValues = false;
 	private List<Input> fieldInputs = new ArrayList<>();
 
+	private Map<String,Long> previousTimestamps = new HashMap<>();
+	private Map<String,Long> previousSequenceNumbers = new HashMap<>();
+
+	private static final Logger log = Logger.getLogger(SendToStream.class);
+
+	private long getNextSequenceNumber(String streamId, long timestamp) {
```
hpihkala: not DRY, see comments in `ModuleWithUI`

src/java/com/unifina/signalpath/kafka/SendToStream.java
```diff
@@ -67,34 +94,47 @@ private void ensureServices() {
 		}
 	}
 
+	private StreamMessage getMsg(Stream stream){
+		int streamPartition = StreamPartitioner.partition(stream, null);
+		long timestamp = System.currentTimeMillis();
+		long sequenceNumber = getNextSequenceNumber(stream.getId(), timestamp);
+		SecUser user = SecUser.getViaJava(getGlobals().getUserId());
```
hpihkala: Same here, can't hit DB for every message

src/java/com/unifina/signalpath/utils/MessageChainUtil.java
```diff
@@ -0,0 +1,73 @@
...
+
+	private SecUser getUser(Long userId) {
+		if (cachedUser == null) {
+			cachedUser = SecUser.getViaJava(userId);
+		}
+		return cachedUser;
+	}
+
+	private long getNextSequenceNumber(String key, long timestamp) {
+		if (!previousTimestamps.containsKey(key) || previousTimestamps.get(key) != timestamp) {
```
hpihkala: Optimization: 1 hashmap lookup instead of 2:
```
Long previousTimestamp = previousTimestamps.get(key);
if (previousTimestamp == null || previousTimestamp != timestamp) {
    return 0L;
}
```

src/java/com/unifina/signalpath/utils/MessageChainUtil.java
```diff
@@ -0,0 +1,73 @@
...
+			return 0L;
+		}
+		if (!previousSequenceNumbers.containsKey(key)) {
```
hpihkala: Optimization: 1 hashmap lookup instead of 2:
```
Long previousSequenceNumber = previousSequenceNumbers.get(key);
return previousSequenceNumber == null ? 0 : previousSequenceNumber + 1;
```
(In this suggestion I also removed the `put`, because `getStreamMessage` does it anyway. Also better pattern not to write anything in methods called `getXXXX`)

src/java/com/unifina/signalpath/utils/MessageChainUtil.java
```diff
@@ -0,0 +1,73 @@
...
+			previousSequenceNumbers.put(key, 0L);
+		}
+		return previousSequenceNumbers.get(key)+1;
+	}
+
+	private MessageRef getPreviousMessageRef(String key) {
+		if (!previousTimestamps.containsKey(key)) {
```
hpihkala: Same here, optimize to 1 lookup instead of 2 by using a variable

src/java/com/unifina/signalpath/utils/MessageChainUtil.java
```diff
@@ -0,0 +1,73 @@
...
+			return null;
+		}
+		if (!previousSequenceNumbers.containsKey(key)) {
+			previousSequenceNumbers.put(key, 0L);
```
hpihkala: - Would be great not to write anything in method `getXXX` (it's not needed here?)
- Is it a valid situation that `previousTimestamp` is found but `previousSequenceNumber` is not? Maybe we should require both to be non-null?

```
Long previousTimestamp = previousTimestamps.get(key);
Long previousSequenceNumber = previousSequenceNumber.get(key);
if (previousTimestamp == null || previousSequenceNumber == null) {
    return null;
} else {
    return new MessageRef(previousTimestamp, previousSequenceNumber);
}

File: src/java/com/unifina/signalpath/utils/MessageChainUtil.java [hunk 26]
```diff
@@ -0,0 +1,73 @@
...
+public class MessageChainUtil implements Serializable {
+	private Map<String,Long> previousTimestamps = new HashMap<>();
+	private Map<String,Long> previousSequenceNumbers = new HashMap<>();
+
+	private transient SecUser cachedUser = null;
+
+	private static final Logger log = Logger.getLogger(MessageChainUtil.class);
+
+	private SecUser getUser(Long userId) {
+		if (cachedUser == null) {
```
hpihkala: I guess we should either:
- Check that `userId` is a match with `cachedUser.getId()`, in case someone is abusing the class, or
- Even better: prevent abusing the class by passing the `userId` in the constructor of this class instead of re-passing it in `getStreamMessage` for every message, since we can't allow the user to change.

test/unit/com/unifina/ModuleTestingSpecification.groovy
```diff
@@ -23,6 +23,7 @@ class ModuleTestingSpecification extends BeanMockingSpecification {
 	}
 
 	protected Globals mockGlobals(Map context=[:], SecUser user = new SecUser(timezone: "UTC")) {
+		user.save(failOnError: true, validate: false)
```
hpihkala: Hmm probably shouldn't add this here because it requires you to pollute tests with `@Mock(SecUser)`, even if the test doesn't need SecUsers to be saved/loaded from db?

You should just `save` the user and `@Mock(SecUser)` it in the tests which load it from the db (such as your new `MessageChainUtilSpec`).

test/unit/com/unifina/signalpath/PropagatorSpec.groovy
```diff
@@ -3,11 +3,14 @@ package com.unifina.signalpath
...
 import com.unifina.data.FeedEvent
 import com.unifina.datasource.DataSource
+import com.unifina.domain.security.SecUser
 import com.unifina.feed.MasterClock
 import com.unifina.signalpath.simplemath.Count
 import com.unifina.signalpath.time.ClockModule
 import com.unifina.utils.Globals
+import grails.test.mixin.Mock
 
+@Mock(SecUser)
```
hpihkala: Yep the change in `ModuleTestingSpecification` adds this pollution. Please revert that change and the added `@Mock`s to unrelated files

src/java/com/unifina/signalpath/utils/MessageChainUtil.java
```diff
@@ -19,43 +19,37 @@
 	private Map<String,Long> previousTimestamps = new HashMap<>();
 	private Map<String,Long> previousSequenceNumbers = new HashMap<>();
 
-	private transient SecUser cachedUser = null;
+	private SecUser user;
```
hpihkala: Re our discussion: should be `transient`

test/unit/com/unifina/signalpath/blockchain/VerifySignatureSpec.groovy
```diff
@@ -1,17 +1,14 @@
...
-import grails.test.mixin.Mock
 
-@Mock(SecUser)
 class VerifySignatureSpec extends ModuleTestingSpecification {
 
 	VerifySignature module
 
 	def setup() {
-		module = setupModule(new VerifySignature())
+		module = setupModule(new VerifySignature(), [:], null)
```
hpihkala: We discussed these changes, add `parentSignalPath.globals = globals` to `setupModule`

hpihkala:
@mthambipillai There are a few places in the frontend code that use the `resend_all` mode for historical canvas runs. That's not going to work with support for `resend_all` dropped in the new versions. One way to work around that would be to replace `resend_all: true` with `resend_from: 0` or similar.

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -170,7 +169,7 @@ class StreamService {
 			}
 
 			int partition = partitioner.partition(stream, null)
-			StreamMessageV30 msg = new StreamMessageV30(stream.id, partition, date.time, sequenceNumber, publisherId,
+			StreamMessageV30 msg = new StreamMessageV30(stream.id, partition, date.time, sequenceNumber, publisherId, "",
```
hpihkala: Let's generate a random `msgChainId` before the `for` loop using `IdGenerator.getShort()` and use it here.

src/java/com/unifina/signalpath/utils/MessageChainUtil.java
```diff
@@ -59,17 +59,12 @@ public StreamMessage getStreamMessage(Stream stream, Date timestampAsDate, Map c
 		long timestamp = timestampAsDate.getTime();
 		long sequenceNumber = getNextSequenceNumber(key, timestamp);
 		String publisherId = getUser().getPublisherId();
-		MessageID msgId = new MessageID(stream.getId(), streamPartition, timestamp, sequenceNumber, publisherId);
+		MessageID msgId = new MessageID(stream.getId(), streamPartition, timestamp, sequenceNumber, publisherId, "");
```
hpihkala: Do `this.msgChainId = IdGenerator.getShort();` in constructor and use it here. `MessageChainUtil` instance is (publisher, msgChainId)-specific. BTW wouldn't hurt to have some class-level comments about the purpose and functionality of this class.

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -169,7 +169,7 @@ class StreamService {
 			}
 
 			int partition = partitioner.partition(stream, null)
-			StreamMessageV30 msg = new StreamMessageV30(stream.id, partition, date.time, sequenceNumber, publisherId, "",
+			StreamMessageV30 msg = new StreamMessageV30(stream.id, partition, date.time, sequenceNumber, publisherId, IdGenerator.getShort(),
```
hpihkala: Generate one `msgChainId` and use that for all messages in the csv import

PR #561, opened 2019-01-31
CORE-1592: Rename integration key API endpoint

kare:


grails-app/conf/UrlMappings.groovy
```diff
@@ -66,7 +66,10 @@ class UrlMappings {
 		"/api/v1/users/me/keys"(resources: "keyApi", excludes: ["create", "edit", "update"]) { resourceClass = SecUser }
 		"/api/v1/users/me/products"(controller: "productApi", action: "index") { operation = Permission.Operation.SHARE }
 
-		"/api/v1/integration_keys"(resources: "integrationKeyApi")
+		"/api/v1/integration_keys"(controller: "integrationKeyApi", action: "index", excludes: ["delete", "update", "save"], method: "GET")
```
hpihkala: What does the `excludes` list accomplish here? Seems like a pain to maintain. I thought it is intended to be used with the `resources` parameter (Grails auto-generated mappings).

grails-app/controllers/com/unifina/controller/api/IntegrationKeyApiController.groovy
```diff
@@ -12,6 +13,13 @@ import groovy.json.JsonSlurper
 
 @Secured(["IS_AUTHENTICATED_ANONYMOUSLY"])
 class IntegrationKeyApiController {
+	static allowedMethods = [
```
hpihkala: Seems like duplication of what you have specified in `UrlMappings.groovy`. Keep this one and remove the method definitions from `UrlMappings`?

grails-app/conf/UrlMappings.groovy
kare: It's the grails magic that is annoying. Now you can see what happens with a single glance.

harbu: I think you shouldn't need `excludes` here since you are mapping only a single API endpoint. Or is it not working properly without it?

grails-app/controllers/com/unifina/controller/api/IntegrationKeyApiController.groovy
```diff
@@ -65,6 +73,25 @@ class IntegrationKeyApiController {
 		render ""
 	}
 
+	@StreamrApi
+	def update(String id) {
+		Map json = new JsonSlurper().parseText((String) request.JSON)
+		if (json.name == null || json.name.trim() == "") {
+			response.status = 400
```
harbu: It would be good to return a JSON error here in the format defined in `com.unifina.api.ApiError`. One way to do it is throw an `com.unifina.api.ApiException` here after which `com.unifina.controller.api.ErrorController` will handle setting the status code and responding with JSON automagically.

So for example `throw new ApiException(400, "BODY_NOT_VALID", "must send JSON body containing property 'name'.")`

grails-app/controllers/com/unifina/controller/api/IntegrationKeyApiController.groovy
```diff
@@ -65,6 +73,25 @@ class IntegrationKeyApiController {
...
+			return
+		}
+		String name = (String) json.name
+
+		try {
+			ethereumIntegrationKeyService.updateKey(apiUser(), id, name)
+		} catch (NotFoundException e) {
+			response.status = 404
```
harbu: No need to catch because `NotFoundException` extends `ApiException` and hence is serialized into a JSON error response (and status code) automagically be ErrorController.

grails-app/conf/UrlMappings.groovy
hpihkala: Removing grails magic is ok. But the `excludes` statement is only needed when you **use** that grails magic. Now you don't need it!

grails-app/controllers/com/unifina/controller/api/IntegrationKeyApiController.groovy
kare: Status codes don't seem to be set by error controller! Which is odd because ErrorControllerSpec seems solid. Perhaps render() sets the response code anyway?

grails-app/conf/UrlMappings.groovy
kare: Ok, sounds good!

PR #563, opened 2019-02-08
CORE-1622: Add Stream status endpoint

kare:


grails-app/controllers/com/unifina/controller/api/StreamApiController.groovy
```diff
@@ -168,4 +172,26 @@ class StreamApiController {
 			action.call(stream)
 		}
 	}
+
+	@StreamrApi
+	def status() {
+		Stream s = Stream.get((String) params.id)
```
harbu: Btw you can have the `id` in the parameter list of the method, e.g., `def status(String id)`. This works for variables defined in the url mapping in `UrlMappings.groovy`. Not saying it needs to be done in that way though, either is fine imo.

grails-app/controllers/com/unifina/controller/api/StreamApiController.groovy
```diff
@@ -168,4 +172,26 @@ class StreamApiController {
...
+		if (s == null) {
+			response.status = 404
```
harbu: Created JIRA issue (CORE-1632) for myself to look into this response status code not being set by exceptions.

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -328,4 +328,24 @@ class StreamService {
...
+			this.date = date
+		}
+	}
+
+	@CompileStatic
+	StreamStatus status(Stream s, Date threshold) {
+		StreamrMessage msg = cassandraService.getLatestFromAllPartitions(s)
+		if (msg == null) {
+			return new StreamStatus(false, null)
+		} else if (msg != null && msg.getTimestamp().before(threshold)) {
```
harbu: Condition `msg != null` is not strictly necessary here. It will always be true because of the if-condition before.

PR #564, opened 2019-02-12
CORE-1593: Remove user account

kare:


grails-app/conf/UrlMappings.groovy
```diff
@@ -62,7 +62,7 @@ class UrlMappings {
 		"/api/v1/modules"(resources: "moduleApi")
 		"/api/v1/modules/$id/help"(controller: "moduleApi", action: "help")
 
-		"/api/v1/users/me"(controller: "userApi", action: "getUserInfo")
```
harbu: Btw have you tested that the `getUserInfo` endpoint works in practice after this URL mapping change? I have a suspicion that you may need to leave in `"/api/v1/users/me"(controller: "userApi", action: "getUserInfo")` because the `(resources: "userApi")` will only map the basic CRUD endpoints (index, create, save, show, edit ... etc. https://grails.github.io/grails2-doc/2.5.6/guide/single.html#restfulMappings) . I may be wrong here, just making sure.

grails-app/services/com/unifina/service/UserService.groovy
```diff
@@ -120,6 +122,17 @@ class UserService {
...
 
+	def delete(SecUser user, Long id) {
+		if (user == null || user.id == null || id == null) {
+			throw new NotFoundException("user not found", "User", null)
+		}
+		if (user.id != id) {
+			throw new ApiException(400, "DELETE_USER_ERROR", "delete user error")
+		}
+		user.enabled = false
+		user.save(validate: false)
```
harbu: why `validate: false` here?

grails-app/services/com/unifina/service/UserService.groovy
```diff
@@ -120,6 +122,17 @@ class UserService {
 		}
 	}
 
+	def delete(SecUser user, Long id) {
+		if (user == null || user.id == null || id == null) {
+			throw new NotFoundException("user not found", "User", null)
+		}
+		if (user.id != id) {
+			throw new ApiException(400, "DELETE_USER_ERROR", "delete user error")
```
harbu: Hmm what about having message "can only delete own user account" or something similar. Allows user of API to more accurately diagnose what the cause was.

PR #567, opened 2019-02-21
CORE-1612 & CORE-1613: Add /api/v1/users/me{/changePassword} REST endpoints

kare:


grails-app/conf/UrlMappings.groovy
```diff
@@ -68,6 +68,9 @@ class UrlMappings {
 		"/api/v1/users/me/keys"(resources: "keyApi", excludes: ["create", "edit", "update"]) { resourceClass = SecUser }
 		"/api/v1/users/me/products"(controller: "productApi", action: "index") { operation = Permission.Operation.SHARE }
 
+		"/api/v1/profile/changePassword"(controller: "profileApi", action: "changePassword", excludes: ["update"])
```
hpihkala: `excludes` unnecessary?

grails-app/controllers/com/unifina/controller/api/ProfileApiController.groovy
```diff
@@ -0,0 +1,85 @@
...
+
+import com.unifina.api.ApiException
+import com.unifina.domain.security.SecUser
+import com.unifina.security.StreamrApi
+import grails.converters.JSON
+import grails.plugin.springsecurity.annotation.Secured
+import grails.validation.Validateable
+
+@Secured(["IS_AUTHENTICATED_ANONYMOUSLY"])
+class ProfileApiController {
```
hpihkala: These user-related actions should be in `UserApiController`?

grails-app/conf/UrlMappings.groovy
hpihkala: These should be under `/api/v1/users/me/...` as that is the prefix which points to the current (logged in) user.

- `POST /api/v1/users/me/changePassword` (special functionality maybe warrants verb endpoint)
- `PUT /api/v1/users/me` for update (REST convention)

test/unit/com/unifina/controller/api/ProfileApiControllerSpec.groovy
```diff
@@ -0,0 +1,175 @@
...
+import com.unifina.api.ApiException
+import com.unifina.api.InvalidUsernameAndPasswordException
+import com.unifina.domain.security.SecUser
+import com.unifina.service.UserService
+import grails.test.mixin.Mock
+import grails.test.mixin.TestFor
+
+@TestFor(ProfileApiController)
+@Mock(SecUser)
+class ProfileApiControllerSpec extends ControllerSpecification {
```
hpihkala: Move to `UserApiControllerSpec`

grails-app/controllers/com/unifina/controller/api/UserApiController.groovy
```diff
@@ -1,29 +1,100 @@
...
+
+	@StreamrApi
+	def changePassword(ChangePasswordCommand cmd) {
+		if (!cmd.validate()) {
+			throw new ApiException(400, "PASSWORD_CHANGE_FAILED", "Password not changed!")
+		}
+		SecUser user = loggedInUser()
+		user.password = springSecurityService.encodePassword(cmd.password)
+		user.save(flush: true, failOnError: true)
+		springSecurityService.reauthenticate(user.username)
```
hpihkala: The new frontend doesn't use the `springSecurityService` cookie-based sessions anymore, so I think this line can (and should) be removed (if the old frontend does not use this endpoint).

PR #569, opened 2019-03-07
CORE-1643: Generate IDs for Stream config fields

kare:


hpihkala:
If these IDs are purely for the frontend, and bear no other significance, shouldn't they be generated in the frontend?

fonty1:
I asked for this so I'll jump in with the backstory:

It could be done on the frontend but it adds some complexity because React requires a unique ID to map these fields in a list. Since we map this list on every render, we cannot generate these dynamically because React will lose focus of the field that's being edited.

Frontend Option 1 
Use the lifecycle hook `getderivedstatefromprops` to mirror the props to the local component state and append a generated ID to each field. This is tricky to implement due to the reordering, live editing and deleting functionality of the list, but doable.

Frontend Option 2
Generate and store the field IDs in redux. We would need to store two separate field lists (one with, one without IDs) so that we have the stream fields in the correct schema for the eventual `PUT /streams/id` endpoint.

Backend Option 3 
Receive an ID from the backend to work with. If it's not best practice then I can make it work with option 1 or 2.

screenshot 2019-03-08 at 14 03 22

hpihkala:
> This is tricky to implement due to the reordering, live editing and deleting functionality of the list, but doable.

Does this become easier if you deterministically derive a key based on the object content? Like Kare is deriving the id by hashing the field name and type. Then React should’t run into internal confusion since the same object will always have the same key?

If this really isn’t easily solved frontend-side, then let’s proceed with these changes. While not intimately familiar with React, I intuitively find it very weird if there’s no trivial solution to this: having an array of objects with no explicit ids must be a very common situation. But I hear you if you say so.

hpihkala:
Aah now I see one issue: the id probably shouldn’t depend on the content. This is a flaw in both my proposal as well as Kare’s implementation. If the if is hash(name, type), the if will change if the name or type change, and React will get confused. Is that what you meant when you said that ”React will lose focus”?

With current implementation this change of ids would also happen on save and (re)load, but I guess this doesn’t produce a UX problem in that situation, as the user isn’t in the middle of editing.

If the above understanding is correct then let’s just proceed with this and get it over with. Better approach would be to generate a random id whenever new rows are added, and persist the id to database in the json, but then we’d have to retroactively add ids to existing stream configs, which might be overkill for the issue at hand.

fonty1:
> If the id is hash(name, type), the id will change if the name or type change, and React will get confused. Is that what you meant when you said that ”React will lose focus”?

Yes, that's right. Any form of randomness or name based hashing causes UX issues with React text editing because we are re-assigning the key on render. 

The best practice in React is to provide a stable and unique identity for each item (the key) from the data. As a last resort they recommend using the array index as the key, but in this case we can't because we allow for reordering. 

>  the id probably shouldn’t depend on the content. This is a flaw in both my proposal as well as Kare’s implementation, but in the backend the flaw doesn’t do as much harm as the change doesn’t happen while the user is interacting.

Yes, I didn't think of that, two fields of the same name and type will cause me more issues (key duplicates). Ideally any static random string will do. The frontend validation could block duplicates but since it has been allowed with the old UI there is probably existing 'tainted' field lists that would be buggy in the new UI.

Juha is facing similar issues in the Editor so it's worthwhile to keep these sorts of React limitations in mind.

fonty1:
Now that I've thought it through I think the simplest path forward here is if I generate a duplicate field field list in the Redux store with unique and stable IDs at the time I receive the field list from the network, then, before the saving `PUT` happens, remove the IDs from the field list.

I can test this out locally but it should work without any work required on the backend.

fonty1:
I've circumvented the ID issue on the frontend so there's no need for any backend modifications.

I will close this PR without merging. Thanks @kare & @hpihkala for the assistance!

PR #570, opened 2019-03-11
CORE-1642: move /module/jsonGetModule and /module/jsonGetModuleTree under API

harbu:
Copy methods of `/module/jsonGetModule` and `/module/jsonGetModuleTree` into `ModuleApiController` and create new URL mappings `/api/v1/modules/jsonGetModule` and `/api/v1/modules/jsonGetModuleTree`.

Tested by hand that works.

### DIFF
--- a/grails-app/conf/UrlMappings.groovy
+++ b/grails-app/conf/UrlMappings.groovy
@@ -62,6 +62,8 @@ class UrlMappings {
 
 		"/api/v1/modules"(resources: "moduleApi")
 		"/api/v1/modules/$id/help"(controller: "moduleApi", action: "help")
+		"/api/v1/modules/jsonGetModule"(controller: "moduleApi", action: "jsonGetModule")
+		"/api/v1/modules/jsonGetModuleTree"(controller: "moduleApi", action: "jsonGetModuleTree")
 
 		"/api/v1/users/me"(controller: "userApi", action: "getUserInfo", excludes: ["delete"])
 		"/api/v1/users/me/$id"(controller: "userApi", action: "delete", excludes: ["getUserInfo"])
--- a/grails-app/conf/UrlMappings.groovy
+++ b/grails-app/conf/UrlMappings.groovy
@@ -62,8 +62,8 @@ class UrlMappings {
 
 		"/api/v1/modules"(resources: "moduleApi")
 		"/api/v1/modules/$id/help"(controller: "moduleApi", action: "help")
-		"/api/v1/modules/jsonGetModule"(controller: "moduleApi", action: "jsonGetModule")
-		"/api/v1/modules/jsonGetModuleTree"(controller: "moduleApi", action: "jsonGetModuleTree")
+		"/api/v1/modules/$id"(controller: "moduleApi", action: "jsonGetModule")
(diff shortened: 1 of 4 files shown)


grails-app/conf/UrlMappings.groovy
```diff
@@ -62,6 +62,8 @@ class UrlMappings {
 
 		"/api/v1/modules"(resources: "moduleApi")
```
hpihkala: On this line, `resources: "moduleApi"` auto-maps the whole CRUD set of endpoints. We don't want that, so we could be more explicit about what is available.

```suggestion
		"/api/v1/modules"(controller: "moduleApi", "action": "index")
```

grails-app/conf/UrlMappings.groovy
```diff
@@ -62,6 +62,8 @@ class UrlMappings {
...
 		"/api/v1/modules/$id/help"(controller: "moduleApi", action: "help")
+		"/api/v1/modules/jsonGetModule"(controller: "moduleApi", action: "jsonGetModule")
```
hpihkala: The endpoints should be renamed to align with RESTful principles. Proposal:

- `jsonGetModule` => `POST /api/v1/modules/$id`

Basically this instantiates a module (identified by `id`) with given config (currently passed as parameter `configuration`). In the renamed endpoint, module `id` is provided in the URL, and what is now passed as `configuration` would simply be the request body.

- Add new endpoint: `GET /api/v1/modules/$id`

Would do the same thing as `POST /api/v1/modules/$id` with empty body (return the module with default config).

- `jsonGetModuleTree` => `GET /api/v1/moduleCategories`

Here we have a few options:

1. Don't include the modules in the list of `moduleCategories` (currently included in `jsonGetModuleTree` under key `children`. This would leave it up to the frontend to piece together the module tree by calling both `/api/v1/moduleCategories` and `GET /api/v1/modules`.

2. Always include the modules in the results (as `jsonGetModuleTree` does currently). Here I propose a key change  `children` => `modules`.

3. (preferred) Make the association fetching optional, e.g. `GET /api/v1/moduleCategories?modules=true` would fetch the association, false by default.

grails-app/conf/UrlMappings.groovy
harbu: Does it make a difference since those methods haven't been defined anyways?

grails-app/conf/UrlMappings.groovy
harbu: Are these changes something you want seen in this PR or could these be done later in a separate refactor PR?

hpihkala: @harbu they aren't big changes, so I was thinking this PR

harbu: I did the `jsonGetModule` remapping. W.r.t. to the module categories I went with `GET /api/v1/module_categories` for consistency with rest of API. I believe the `children` may also contain sub-categories not just modules?

harbu: Swagger also updated.

hpihkala: > I believe the children may also contain sub-categories not just modules?

True, good point.

PR #573, opened 2019-03-18
CORE-1650: Fix URL mappings for user delete/update/info

kare:
Current mappings are:
- GET "/api/v1/users/me"
- PUT "/api/v1/users/me/update"
- DELETE "/api/v1/users/me/delete"


hpihkala:
@kare @harbu Now the urls are in conflict with rest best practices and docs. Could this have been fixed with adding the appropriate `method` to the `UrlMappings`? For example:

```
"/api/v1/users/me"(method: "GET", controller: "userApi", action: "getUserInfo")
"/api/v1/users/me"(method: "PUT", controller: "userApi", action: "update")
"/api/v1/users/me"(method: "DELETE", controller: "userApi", action: "delete")
```

kare:
@hpihkala @harbu I'll test it

PR #575, opened 2019-03-20
CORE-1647: Example resources for Stream and Canvas

kare:


grails-app/services/com/unifina/service/CanvasService.groovy
```diff
@@ -201,8 +202,50 @@ class CanvasService {
...
 	}
 
+	@CompileStatic
+	def addExampleCanvases(SecUser user, List<Canvas> examples) {
+		for (Canvas c : examples) {
+			switch (c.exampleType) {
+				// Create a copy of the example canvas for the user and grant all permissions.
+				case ExampleType.COPY:
+					Canvas canvas = new Canvas()
+					canvas.dateCreated = c.dateCreated
```
hpihkala: This code will be difficult to maintain if we enumerate all the fields explicitly. In groovy you can do something like `canvas.properties = c.properties` to do it via reflection. Would that work for copying most of the fields? Then you can overwrite the necessary fields explicitly (like `canvas.exampleType = ExampleType.NOT_SET`)

grails-app/services/com/unifina/service/CanvasService.groovy
```diff
@@ -201,8 +202,50 @@ class CanvasService {
...
+					canvas.lastUpdated = c.lastUpdated
+					canvas.name = c.name
+					canvas.json = c.json
```
hpihkala: We should write `canvas.json` **after** the `resetUiChannels`, reconstruction, and `ensureUiChannel` calls

grails-app/services/com/unifina/service/CanvasService.groovy
```diff
@@ -201,8 +202,50 @@ class CanvasService {
...
+					canvas.requestUrl = null //c.requestUrl
+					canvas.serialization = null //c.serialization
+					canvas.startedBy = null //c.startedBy
+					canvas.exampleType = ExampleType.NOT_SET
+					c.save(validate: true, failOnError: true)
+					permissionService.systemGrant(user, canvas, Permission.Operation.SHARE)
+
+					Map signalPathMap = (Map) JSON.parse(canvas.json)
+					resetUiChannels(signalPathMap)
+					SignalPathService.ReconstructedResult result = reconstructFrom(signalPathMap, user)
```
hpihkala: Much of the code here is now copied from elsewhere in `CanvasService`. Can't we use `canvasService.createNew(...)` to duplicate the example canvas? This function would automatically reset the ui channels, reconstruct, grant permission, and ensure the UI channels exist. You wouldn't have to duplicate the code here.

grails-app/services/com/unifina/service/CanvasService.groovy
```diff
@@ -201,8 +202,50 @@ class CanvasService {
...
+					break
+				// Grant read permission to example canvas.
+				case ExampleType.SHARE:
+					permissionService.systemGrant(user, c, Permission.Operation.READ)
+					break
+			}
+		}
+	}
+
 	private boolean hasCanvasPermission(Canvas canvas, SecUser user, Permission.Operation op) {
```
hpihkala: One-liner private function seems useless now. Could just as well call `permissionService.check` directly in the places where this function is used.

grails-app/domain/com/unifina/domain/signalpath/Canvas.groovy
```diff
@@ -67,10 +70,10 @@ class Canvas {
 		id generator: IdGenerator.name // Note: doesn't apply in unit tests
 		json type: 'text'
 		hasExports defaultValue: false
-		example defaultValue: false
 		adhoc defaultValue: false
 		runner index: 'runner_idx'
 		dashboardItems cascade: 'all-delete-orphan'
+		exampleType enumType: "identity", defaultValue: ExampleType.NOT_SET
```
hpihkala: Queries to fetch the examples can become slow if there are many rows. Should we add an index for this field?

grails-app/domain/com/unifina/domain/data/Stream.groovy
```diff
@@ -62,6 +66,7 @@ class Stream implements Comparable {
 		requireSignedData defaultValue: "false"
 		autoConfigure defaultValue: "true"
 		storageDays defaultValue: DEFAULT_STORAGE_DAYS
+		exampleType enumType: "identity", defaultValue: ExampleType.NOT_SET
```
hpihkala: Queries to fetch the examples can become slow if there are many rows. Should we add an index for this field?

grails-app/services/com/unifina/service/CanvasService.groovy
kare: I'd like to leave it like that because it is more readable.

grails-app/domain/com/unifina/domain/signalpath/Canvas.groovy
kare: Excellent point!

grails-app/domain/com/unifina/domain/data/Stream.groovy
kare: Excellent point!

grails-app/services/com/unifina/service/CanvasService.groovy
```diff
@@ -204,41 +205,38 @@ class CanvasService {
...
 					SignalPathService.ReconstructedResult result = reconstructFrom(signalPathMap, user)
-					result.signalPath.setCanvas(canvas)
+					result.signalPath.setCanvas(c)
 					result.signalPath.getModules().each {
 						if (it instanceof ModuleWithUI) {
 							it.ensureUiChannel()
 						}
 					}
 					result.signalPath.ensureUiChannel()
+					c.json = example.json
```
hpihkala: We should set the canvas json to the reconstructed one (basically `c.json = result.map as JSON`).

grails-app/services/com/unifina/service/CanvasService.groovy
```diff
@@ -204,41 +205,38 @@ class CanvasService {
...
+					c.runner = null
+					c.server = null
+					c.requestUrl = null
+					c.serialization = null
+					c.startedBy = null
+					c.state = Canvas.State.STOPPED
+					c.exampleType = ExampleType.NOT_SET
 					c.save(validate: true, failOnError: true)
-					permissionService.systemGrant(user, canvas, Permission.Operation.SHARE)
+					permissionService.systemGrant(user, c, Permission.Operation.SHARE)
```
hpihkala: I think we actually need to grant all of `READ`, `WRITE` and `SHARE` explicitly.

grails-app/services/com/unifina/service/CanvasService.groovy
hpihkala: @kare how about this?

grails-app/services/com/unifina/service/CanvasService.groovy
hpihkala: Ok, no problem

grails-app/services/com/unifina/service/CanvasService.groovy
kare: done

grails-app/services/com/unifina/service/CanvasService.groovy
```diff
@@ -211,6 +211,8 @@ class CanvasService {
 				case ExampleType.COPY:
 					Canvas c = new Canvas()
 					InvokerHelper.setProperties(c, example.properties)
+					c.dashboardItems = new HashSet<DashboardItem>()
+					c.permissions = new HashSet<Permission>()
```
kare: InvokeHelper creates some funny references and hibernate dies. Here is a fix, however I cannot remember if Hibernate has a problem with this? It is tested and works with example canvas copy.

grails-app/services/com/unifina/service/CanvasService.groovy
```diff
@@ -201,8 +203,47 @@ class CanvasService {
...
+					c.runner = null
+					c.server = null
+					c.requestUrl = null
+					c.serialization = null
+					c.startedBy = null
+					c.state = Canvas.State.STOPPED
+					c.exampleType = ExampleType.NOT_SET
+					c.save(validate: true, failOnError: true)
```
kare: Should save be called after updateExisting(...)?

kare: Or is calling save() here useful at all?

hpihkala: `updateExisting` saves the canvas AND grants the user all permissions to it if it is a previously unsaved canvas. So for `ExampleType.COPY` it should be enough to setup the `Canvas` object and `SaveCanvasCommand` object, and call `updateExisting` - nothing else.

grails-app/services/com/unifina/service/CanvasService.groovy
hpihkala: This is a bit of a time bomb... If we add new one-to-many associations to `Canvas`, this will break again (and prevent new users from signing up!). 

The problem might be that `InvokerHelper` is dumb enough to try and copy the associations of `example` to the new `Canvas`, which of course can't work at all.

Have you tried `Canvas c = new Canvas(example.properties)`? This shouldn't try to bind unbindable properties (like `id`), but I'm not sure how it handles those associative `Set`s.

grails-app/services/com/unifina/service/CanvasService.groovy
```diff
@@ -220,19 +222,17 @@ class CanvasService {
...
+					permissionService.systemGrant(user, c, Permission.Operation.WRITE)
+
+					Map map = (JSONObject) JSON.parse(example.json)
+					SaveCanvasCommand cmd = new SaveCanvasCommand(
+						name: c.name,
+						modules: map?.getJSONArray("modules") == null ? [] : map?.getJSONArray("modules"),
+						settings: map?.getJSONObject("settings") == null ? [:] : map?.getJSONObject("settings"),
+						adhoc: false,
+					)
+					updateExisting(c, cmd, user,true)
```
hpihkala: `updateExisting` saves the canvas AND grants the user all permissions to it, if it is a previously unsaved canvas. So for `ExampleType.COPY` it should be enough to:

- setup the `Canvas` object
- setup the `SaveCanvasCommand` object
- `updateExisting`
- nothing else.

So I think you don't need the `c.save(...)` and `permissionService.systemGrant(...)`s.

grails-app/services/com/unifina/service/UserService.groovy
```diff
@@ -69,6 +73,17 @@ class UserService {
 			// Transfer permissions that were attached to sign-up invitation before user existed
 			permissionService.transferInvitePermissionsTo(user)
 		}
+
+		List<Canvas> canvasExamples = Canvas.createCriteria().list {
```
hpihkala: One afterthought: should we wrap these in `try-catch` to make 100% sure that new user creation doesn't fail if there are any exceptions thrown while copying the examples? Maybe there should be a `try-catch` for canvases and streams separately, so that one of them can succeed if one fails.

PR #576, opened 2019-03-21
fix: GET/PUT/DELETE /api/v1/users/me

kare:
Adding method to UrlMappings enables several HTTP methods for one endpoint. Current mappings are:

- GET "/api/v1/users/me"
- PUT "/api/v1/users/me"
- DELETE "/api/v1/users/me"


kare:
Ok! CORE-1656

PR #577, opened 2019-03-21
Fix CORE-1652 streamr-web3 array return value from smart contract

jtakalai:
Not sure about the fix, since I was unable to successfully run a canvas in the engine-and-editor


src/java/com/unifina/signalpath/blockchain/EthereumToStreamrTypes.java
```diff
@@ -74,4 +76,19 @@ public static Output asOutput(String type, String name, AbstractSignalPathModule
 			return new Output(owner, name, "Object");
 		}
 	}
+
+	public static void convertAndSend(Output output, JsonElement value) {
+		if (output instanceof ListOutput) {
+			ArrayList val = new Gson().fromJson(value, ArrayList.class);
+			output.send(val);
```
jtakalai: These lines 80...83 are the gist of this commit that fixes the issue, probably. Everything else is refactoring fluff.

jtakalai: So in case there's something terminally wrong with the rest, these lines can and should probably be salvaged.

PR #579, opened 2019-03-22
replace murmur3 with md5

mthambipillai:
Replace the murmur3 hash function with md5 when converting `partitionKey` to `streamPartition` to reflect the same changes made in the JS and Java clients.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,6 +1,6 @@
 <component name="ProjectCodeStyleConfiguration">
   <code_scheme name="Project" version="173">
-    <JSCodeStyleSettings version="0">
+    <JSCodeStyleSettings>
       <option name="USE_SEMICOLON_AFTER_STATEMENT" value="false" />
       <option name="FORCE_SEMICOLON_STYLE" value="true" />
       <option name="USE_DOUBLE_QUOTES" value="false" />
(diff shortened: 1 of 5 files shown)


.idea/codeStyles/Project.xml
```diff
@@ -1,6 +1,6 @@
 <component name="ProjectCodeStyleConfiguration">
   <code_scheme name="Project" version="173">
-    <JSCodeStyleSettings version="0">
```
kare: What version of IDEA you are running? I've got 2018.3.5

.idea/misc.xml
```diff
@@ -3,7 +3,7 @@
   <component name="JavaScriptSettings">
     <option name="languageLevel" value="FLOW" />
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="false" project-jdk-name="1.8.0.171-oracle" project-jdk-type="JavaSDK">
```
kare: This is the production JDK version. I'd like to keep it like that!

.idea/codeStyles/Project.xml
mthambipillai: I have 2018.2.4

kare: Please upgrade!

PR #583, opened 2019-04-02
Inbox streams

mthambipillai:
Inbox streams are used for direct communications between publishers and subscribers (needed for cryptographic key exchanges).
This PR includes:
- A new boolean field 'inbox' for every stream. Default is false.
- On Ethereum user creation, create an inbox stream for that user (stream.id = user.username, stream.inbox = true).
- When granting an Ethereum subscriber READ permission to a stream, grant him WRITE permission to the inbox streams of the publishers of that stream.
- When granting an Ethereum subscriber READ permission to a stream, grant every publisher of that stream WRITE permission to the subscriber's inbox stream.


test/unit/com/unifina/service/EthereumIntegrationKeyServiceSpec.groovy
```diff
@@ -267,6 +268,16 @@ class EthereumIntegrationKeyServiceSpec extends Specification {
 		SecUser.count == 1
 	}
 
+	void "createEthereumUser() creates inbox stream"() {
```
hpihkala: Add tests:
- Adding Ethereum ID should create inbox stream
- Adding Ethereum private key (`IntegrationKey` of type `Ethereum`) should create inbox stream

grails-app/services/com/unifina/service/EthereumIntegrationKeyService.groovy
```diff
@@ -81,6 +82,8 @@ class EthereumIntegrationKeyService {
 			] as JSON).toString()
 		).save(flush: true)
 
+		createUserInboxStream(address)
```
hpihkala: Should also be done in `createEthereumAccount`

test/unit/com/unifina/service/PermissionServiceSpec.groovy
```diff
@@ -66,6 +72,21 @@ class PermissionServiceSpec extends Specification {
 		dashReadPermission = service.grant(anotherUser, dashAllowed, me, Operation.READ)
 		dashAnonymousReadPermission = service.grantAnonymousAccess(anotherUser, dashPublic)
 		service.grant(anotherUser, dashAllowed, anonymousKey)
+
+		streamService = Mock(StreamService)
+		ethereumIntegrationKeyService = Mock(EthereumIntegrationKeyService)
+
+		// Setup application context
+		def applicationContext = Stub(ApplicationContext) {
```
hpihkala: This test should extend `BeanMockingSpecification` and use the provided utility methods to register the Mocks as actual beans instead of mocking `ApplicationContext` and `GrailsApplication`

test/unit/com/unifina/service/PermissionServiceSpec.groovy
```diff
@@ -201,6 +222,39 @@ class PermissionServiceSpec extends Specification {
 		service.getPermissionsTo(dashPublic, anonymousKey)[0].operation == Operation.READ
 	}
 
+	void "systemGrant() on an Ethereum user and a stream creates also inbox permissions"() {
```
hpihkala: Tests for permission revokes and changes in publisher set are also needed

test/unit/com/unifina/service/SubscriptionServiceSpec.groovy
```diff
@@ -17,16 +17,28 @@ import spock.lang.Specification
...
 
 	void setup() {
 		user = new SecUser(username: "me@streamr.com").save(failOnError: true, validate: false)
 		user2 = new SecUser(username: "someoneElse@streamr.com").save(failOnError: true, validate: false)
+		user3 = new SecUser(username: "0x26e1ae3f5efe8a01eca8c2e9d3c32702cf4bead6").save(failOnError: true, validate: false)
 		s1 = new Stream(name: "stream-1")
 		s2 = new Stream(name: "stream-2")
 		s3 = new Stream(name: "stream-3")
+
+		pub1Inbox = new Stream(name: "publisher1", inbox: true)
```
hpihkala: Lots of new declarations but no new assertions? 

(This class shouldn't actually even be testing/concerned with anything that happens inside `PermissionService`. Looks like this test is using the real `PermissionService`, while it should be testing against a mock. Easy refactor?)

grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -265,17 +268,46 @@ class PermissionService {
...
 	Permission systemGrant(Userish target, resource, Operation operation=Operation.READ) {
+		return systemGrant(target, resource, operation, null, null)
+	}
+
+	Permission systemGrant(Userish target, resource, Operation operation=Operation.READ, Subscription subscription, Date endsAt) {
 		target = target.resolveToUserish()
 		String userProp = getUserPropertyName(target)
 		String resourceProp = getResourcePropertyName(resource)
 
+		if (userProp == "user" && resourceProp == "stream" && operation == Operation.READ) {
```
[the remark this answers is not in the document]
mthambipillai: The `grantInboxStreamPermissions()` method anyway calls methods in the `StreamService` using the bean. So could we simply have this whole method in the `StreamService` and call it in the same way using:
```
StreamService streamService = grailsApplication.mainContext.getBean(StreamService)
```

test/unit/com/unifina/service/SubscriptionServiceSpec.groovy
mthambipillai: I tried but the refactor is not that easy as most of these tests assert on the creation / update / deletion of `Permission`s instead of asserting calls of a mock `PermissionService`. Since this is prior to this PR and a different issue, I feel like it should be done in a separate PR.

grails-app/domain/com/unifina/domain/security/Permission.groovy
```diff
@@ -57,6 +57,8 @@ class Permission {
 	/** When does this Permission expire? null == forever valid */
 	Date endsAt
 
+	Permission parent
```
hpihkala: Should there be some sort of cascade rule? What we want is that if the `parent` gets deleted, then the children get deleted automatically (instead of failing with a foreign key constraint error). Best case is that this happens on the database level. So we should define it in the domain object, and if Grails is smart we'll se a cascade delete in the generated foreign key in the migration file. If Grails is not that smart then Grails will probably attempt to handle it. But the [definition exists](https://grails.github.io/grails2-doc/2.5.x/ref/Database%20Mapping/cascade.html) for sure.

hpihkala: Oh I noticed now that child permissions are being explicitly deleted in the code when permissions are revoked. That should work, although leveraging the built-in cascade functionality may be cleaner (possibly more unstable too in case of Grails, haha).

grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -502,10 +490,15 @@ class PermissionService {
 				perm.operation == op
 			}.each { Permission perm ->
 				revoked.add(perm)
+				List<Permission> childPermissions = Permission.findAllByParent(perm)
+				revoked.addAll(childPermissions)
 				try {
 					log.info("performRevoke: Trying to delete permission $perm.id")
 					Permission.withNewTransaction {
 						perm.delete(flush: true)
+						for (Permission childPerm: childPermissions) {
```
hpihkala: Child permissions need to be deleted first before deleting the parent permission. (It might not matter here in which way you write the code, as it happens in the same transaction. If this code works, then that must be the case.)

grails-app/services/com/unifina/service/PermissionService.groovy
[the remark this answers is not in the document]
hpihkala: This ties back to @harbu 's [comment](https://github.com/streamr-dev/engine-and-editor/pull/583#discussion_r279562061). For me, the current proposal actually made it sort of worse:

Original situation:
- Inbox-specific stuff in `PermissionService`

Current situation
- Inbox-specific stuff in `PermissionService`
- Circular dependency between `StreamService` and `PermissionService`
- Creation of `Permission`s in `StreamService` (wrong place)

If these can't be cleanly decoupled via some event system as suggested above, then it's probably best to just go with the original problem of "Inbox-specific stuff in `PermissionService`". Maybe add a comment there to try and fix it later in Grails 3.

grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -265,15 +268,92 @@ class PermissionService {
...
 			(resourceProp): resource,
 			(userProp): target,
 			operation: operation,
+			subscription: subscription,
+			endsAt: endsAt
 		).save(flush: true, failOnError: true)
+
+		// When a user is granted read access (subscriber) or write access (publisher) to a stream,
+		// we need to set the corresponding inbox stream permissions (see methods comments below).
+		if (userProp == "user" && resourceProp == "stream") {
```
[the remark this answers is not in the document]
hpihkala: I agree with the comment. See similar discussion [here](https://github.com/streamr-dev/engine-and-editor/pull/583#discussion_r275511715), with a suggested solution but unfortunately it might not be feasible for this version of Grails.

grails-app/services/com/unifina/service/PermissionService.groovy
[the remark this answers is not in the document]
hpihkala: There is actually something called [GORM events](https://grails.github.io/grails2-doc/2.5.x/guide/single.html#eventsAutoTimestamping). We could tap into the `afterInsert`/`beforeDelete` events for `Permission` objects to create/delete the child permissions. That won't be 100% decoupled either as we'd define the handlers (which are inbox-stream-specific) in `Permission.groovy`, but it seems cleaner than the current impl.

test/unit/com/unifina/controller/api/StreamApiControllerSpec.groovy
```diff
@@ -57,6 +60,12 @@ class StreamApiControllerSpec extends ControllerSpecification {
 		streamFourId = streamService.createStream([name: "otherUserStream", feed: feed], otherUser).id
 
 		controller.streamService = streamService
+
+		// Setup grailsApplication
+		def grailsApplication = new DefaultGrailsApplication()
```
hpihkala: These two lines may be unnecessary, as `grailsApplication` should be available in the unit test scope? I'm not sure about this.

grails-app/domain/com/unifina/domain/security/Permission.groovy
mthambipillai: I didn't try the built-in Grails feature since this code works and looks simple and clean enough to me. I personally think it's better to have this simple logic in explicit viewable code instead of relying on some under the hood and potentially unstable Grails mechanism.

grails-app/services/com/unifina/service/PermissionService.groovy
mthambipillai: Yes this code works since it's the same transaction.

grails-app/services/com/unifina/service/PermissionService.groovy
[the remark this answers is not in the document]
harbu: I think the original solution (at least the version which I read) also had the circular dependency problem. It was under a private helper method.

In one way the original solution is the most cleanest if you think about it from the perspective of someone trying to understand / debug exactly what the method `systemGrant` does. There is less jumping around code and what the method does is laid down quite straightforward.

The perspective I had was from someone reading the code from a mindset of "how do inbox streams work?". Having all inbox stream functionality lumped into one class does aid in that. However, it does hamper readability of someone trying to understand what exactly `systemGrant` does because there is more jumping between methods involved. 

The event emitting solution has its own pros/cons. It decouples the PermissionService from inbox stream functionality nicely. What feels a bit off about it is the inbox stream functionality being defined in the `Permission.groovy` class. I think it could potentially make it hard to find / surprise a developer with its behavior. Another concern I have is unit tests: are events emitted correctly in unit tests? (GORM support in unit tests is known to be rather spotty) . The latter concern can be tested quickly ofc.

Maybe a more flexible event system would work best, but I don't think implementing a custom Observer pattern on PermissionService makes sense in this one case.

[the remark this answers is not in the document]
harbu: I think the original solution is probably the best in this case tbh.

PR #588, opened 2019-04-10
CORE-1681: PUT Stream requireSignedData, storageDays and autoConfigure fields

kare:


fonty1:
For context, these are stream settings that can be edited by the user, and thus the PUT inclusion. 
I'd be curious for the default values for these too.

Screenshot 2019-04-10 at 17 16 27
Screenshot 2019-04-10 at 17 16 19

kare:
```
	public final static Integer DEFAULT_STORAGE_DAYS = 365
	Boolean requireSignedData = false
	Boolean autoConfigure = true
```

grails-app/controllers/com/unifina/controller/api/StreamApiController.groovy
```diff
@@ -66,6 +66,9 @@ class StreamApiController {
 			stream.name = newStream.name
 			stream.description = newStream.description
 			stream.config = readConfig()
+			stream.autoConfigure = newStream.autoConfigure
```
hpihkala: Should we only overwrite these fields if they are present? Otherwise we risk overwriting them with `null`
```suggestion
			stream.autoConfigure = newStream.autoConfigure != null ? newStream.autoConfigure : stream.autoConfigure
```

kare: For some reason I tend to forget about 'zero value'.

PR #593, opened 2019-04-30
CORE-1593: Delete user endpoint

kare:


kare:
What about cascades?

kare:
@hpihkala @harbu Should I delete canvases and streams?

hpihkala:
Guys I don't think this is a good approach. For now, "deleting" user accounts should just disable them, as planned.

If that isn't working for some reason, then **that's** a bug that should be fixed.

hpihkala:
Logging in with a disabled user will be fixed in #596. I'm closing this PR in favor of the original plan of simply disabling a user, because it is a far less problematic approach for now. If we just disable the user, we don't need to solve the complexity in deleting their resources - something which can easily lead to orphaned resources or disrupt other users' activities in case of shared resources.

PR #596, opened 2019-05-06
return 401 when signing a challenge with a disabled user's ethereum ac…

mthambipillai:
…count


hpihkala:
@kare do you agree with the above reasoning? I would guess so, as in your PR #597 you had opted to place the check in the controller, which I thought was the right thing to do in this case. 

However this PR has other advantages in that it covers all login methods, and throws named exceptions and cleanly maps them to a 401 in the error handler. So I propose we continue with this PR with @mthambipillai moving the check to `LoginApiController` and the tests to `LoginApiControllerSpec` in the spirit of #597.

kare:
@hpihkala @mthambipillai Yes, I approve and accept these changes.

PR #597, opened 2019-05-06
CORE-1704: Check SecUser.enabled attribute on login

kare:


hpihkala:
Closing this as duplicate of #596. I think this PR has the correct approach of placing the check in the controller, but the other PR checks all the different login methods, and is a bit cleaner in that it properly throws a named exception and maps it to 401 status code in the error handler. Let's work on #596 to get best parts of both submissions.

PR #598, opened 2019-05-09
CORE-1365: Add Stream delete{AllData,DataUpTo,DataRange} API

kare:


grails-app/controllers/com/unifina/controller/api/StreamApiController.groovy
```diff
@@ -203,4 +204,31 @@ class StreamApiController {
 			date: iso8601.format(status.date),
 		] as JSON)
 	}
+
+	@StreamrApi
+	def deleteDataUpTo(String id) {
+		getAuthorizedStream(id, Operation.WRITE) { Stream stream ->
+			Date date = iso8601cal.parse((String) params.date)
```
harbu: Handle case where parameter `date` is not provided? E.g. 400 error and typical JSON error payload.

grails-app/controllers/com/unifina/controller/api/StreamApiController.groovy
```diff
@@ -203,4 +204,31 @@ class StreamApiController {
...
+		getAuthorizedStream(id, Operation.WRITE) { Stream stream ->
+			streamService.deleteAllData(stream)
+			render(status: 204)
+		}
+	}
+
+	@StreamrApi
+	def deleteDataRange(String id) {
+		getAuthorizedStream(id, Operation.WRITE) { Stream stream ->
+			Date start = iso8601cal.parse((String) params.start)
```
harbu: Handle case where parameter `start` or `end` is not provided?

grails-app/controllers/com/unifina/controller/api/StreamApiController.groovy
```diff
@@ -25,6 +26,7 @@ class StreamApiController {
 	]
 
 	private final SimpleDateFormat iso8601 = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'")
+	private final SimpleDateFormat iso8601cal = new SimpleDateFormat("yyyy-MM-dd")
```
hpihkala: We could use this opportunity to make the API better. With dates only, the timezone remains ambiguous (based on server timezone), and there is no way to give a more precise timestamp (including time of day).

In our other API endpoints that accept dates (eg. range resend requests in `data-api`) the API accepts two time formats:
- milliseconds since epoch
- ISO8601 strings, as on line 28 here

Let's use this logic for these endpoints too!

So the helper function would look something like this:

```
private Date parseDate(String input) {
    try {
        // try milliseconds timestamp
        return new Date(Long.parseLong(input));
    } catch (...) {
        // fallback to iso8601 timestamp
        return iso8601.parse(input);
    }
}
```

Please add tests that pass both kinds of timestamps!

If date-only strings like `2019-05-10` won't work anymore, that's ok. The frontend guys can update to adjust. Better to have a unified API.

kare: url parameter dates (date, end, start) are now in format:
2019-05-10T11:45:40Z
or
1557488802759 -> milliseconds from epoch

PR #599, opened 2019-05-13
CORE-1713: GET endpoint for detectFields without save

kare:


grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -101,15 +101,17 @@ class StreamService {
...
 		if (fields) {
-			Map config = stream.getStreamConfigAsMap()
-			config.fields = fields
-			stream.config = gson.toJson(config)
-			stream.save(flush: true, failOnError: true)
+			if (saveFields) {
+				Map config = stream.getStreamConfigAsMap()
+				config.fields = fields
+				stream.config = gson.toJson(config)
+				stream.save(flush: true, failOnError: true)
```
hpihkala: Let's make it so that this `stream.save()` call is the only line depending on `saveFields`. We want to set the `stream.config` because the `StreamApiController` renders `stream.toMap()` as the result.

Potential caveat: Grails might automagically `save` dirty domain objects even if we don't explicitly call `stream.save()`. Would be good to test by hand and verify whether this happens or not. We want it to **not** happen.

A unit test would be nice too.

PR #602, opened 2019-05-15
Retry transaction receipt

hpihkala:
Handle the situation where a tx receipt is not immediately available after submitting a transaction to an Ethereum node to deploy a contract. Old code often produced NPE. New code retries periodically until tx receipt is non-null or max retries is reached.

There should be a test for this, but I don't have time to add it right now. @isentropy please feel free to contribute a test.

If the beta launch draws nearer and we want to merge this without adding a test, we can create a technical debt ticket to add the test later.

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -266,21 +266,21 @@ environments {
  * Streamr-web3 Ethereum bridge address
  */
 streamr.ethereum.defaultNetwork = "rinkeby"
-streamr.ethereum.networks = System.getProperty("streamr.ethereum.networks") ? new Gson().fromJson(System.getProperty("streamr.ethereum.networks")) : [
+streamr.ethereum.networks = System.getProperty("streamr.ethereum.networks") ? new Gson().fromJson(System.getProperty("streamr.ethereum.networks"), Map) : [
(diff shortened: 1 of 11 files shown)


isentropy:
@hpihkala let me refactor this and the method in SendEthereumTransaction to a static helper method:

static TransactionReceipt waitForTransactionReceipt(String txHash, long waitMsBetweenTries, int tries)

I have laptop with me at hospital now.

isentropy:
@hpihkala I have some time to work on this now

isentropy:
@hpihkala I added waitForTransactionReceipt() static method and removed getTransactionReceipt()(which didn't retry). GetEvents uses waitForTransactionReceipt() now too. I think it's good to go now.

hpihkala:
Did some additional cleanup which will simplify configuring the Ethereum networks via system properties, as well as remove the overlap/conflict between items set via `streamr.ethereum.networks` (used by the old Ethereum modules) and `streamr.ethereum.rpcUrls` (used by the new ones) - and both used in `EthereumModuleOptions`.

The changes will break the old deprecated Ethereum modules, but they are broken anyway due to the Ethereum nodes they use being broken, so no real harm done. Best to have clean code and things working going forward. Production backend will begin using Infura nodes for the time being, until we find something better.

PR #604, opened 2019-05-21
CORE-1719 use PENDING block to compute nonce, GetEvents bugfix

isentropy:
- uses PENDING block to compute nonce. fixes the multiple transaction issue
- I added a small fix to Web3jHelper that allows "address" type to be instantiated from BigInteger and Uint160 in addition to String. This was in my PR to web3j, but didn't make it into web3j PR to E&E

### DIFF
--- a/src/java/com/unifina/signalpath/blockchain/GetEvents.java
+++ b/src/java/com/unifina/signalpath/blockchain/GetEvents.java
@@ -1,9 +1,7 @@
 package com.unifina.signalpath.blockchain;
 
-import com.mashape.unirest.http.exceptions.UnirestException;
 import com.streamr.client.protocol.message_layer.ITimestamped;
 import com.unifina.data.FeedEvent;
-import com.unifina.datasource.DataSource;
 import com.unifina.datasource.IStartListener;
 import com.unifina.datasource.IStopListener;
 import com.unifina.signalpath.*;
@@ -11,7 +9,6 @@ import org.apache.log4j.Logger;
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.web3j.abi.EventValues;
-import org.web3j.abi.TypeReference;
 import org.web3j.abi.datatypes.Event;
 import org.web3j.protocol.Web3j;
 import org.web3j.protocol.core.methods.response.Log;
@@ -153,8 +150,16 @@ public class GetEvents extends AbstractSignalPathModule implements ContractEvent
 					log.error("Couldnt find TransactionReceipt for transaction " + txHash + " in allotted time");
 					return;
 				}
-				long ts = Web3jHelper.getTime(web3j, txr);
-				enqueueEvent(new LogsResult(txHash, new Date(ts), txr.getLogs()));
+				long blockts = Web3jHelper.getBlockTime(web3j, txr);
+				Date ts;
+				if(blockts < 0){
+					ts = getGlobals().time;
+					log.error("No block timestamp found for txHash "+txHash+ ". Using globals timestamp "+ts);
+				}
+				else{
+					ts = new Date(blockts*1000);
+				}
+				enqueueEvent(new LogsResult(txHash, ts, txr.getLogs()));
 			}
 		} catch (JSONException | IOException e) {
 			onError(e.getMessage());
--- a/src/java/com/unifina/signalpath/blockchain/GetEvents.java
+++ b/src/java/com/unifina/signalpath/blockchain/GetEvents.java
@@ -1,6 +1,9 @@
 package com.unifina.signalpath.blockchain;
 
 import com.mashape.unirest.http.exceptions.UnirestException;
+import com.streamr.client.protocol.message_layer.ITimestamped;
+import com.unifina.data.FeedEvent;
+import com.unifina.datasource.DataSource;
 import com.unifina.datasource.IStartListener;
 import com.unifina.datasource.IStopListener;
 import com.unifina.signalpath.*;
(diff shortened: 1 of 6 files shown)


isentropy:
fixed a bug in GetEvents: it was only looking at the first event from JSON RPC response for some reason. It was a relic from the old GetEvents code.

src/java/com/unifina/signalpath/blockchain/GetEvents.java
```diff
@@ -103,17 +103,19 @@ static void convertAndSend(Output output, Object value) {
...
+				String txHash = events.getJSONObject(i).getString("transactionHash");
+				txHashOutput.send(txHash);
+				log.info(String.format("Received event '%s'", txHash));
+				TransactionReceipt txr = Web3jHelper.waitForTransactionReceipt(web3j, txHash, check_result_waitms, check_result_max_tries);
+				if (txr == null) {
+					log.error("Couldnt find TransactionReceipt for transaction " + txHash + " in allotted time");
+					return;
+				}
+				displayEventsFromLogs(txr.getLogs());
+				asyncPropagator.propagate();
```
isentropy: @hpihkala @jtakalai  I don't exactly understand this asyncPropagator.propagate() call. I think Juuso added it in the stub he gave me. I'm still having an issue propagating all events from a single JSON response. When I run this normally, only the 1st value is propagated, but when I put a breakpoint here (line 117) and wait a few seconds, it propagates.

hpihkala: @isentropy I think it's because the events aren't being added to the event queue and handled (and propagated) in the event queue consumer thread. How this is done is a bit messy currently, but it will actually become much cleaner in a refactor which is almost ready. Anyway, the current pattern is this:

- In the method called by your module-specific thread (here, `onEvent`), produce events to the DataSource event queue. Best pattern would probably be insert one `FeedEvent` per each event in the transaction.

```
DataSource dataSource = getGlobals().getDataSource();
for (/* each event in transaction */) {
    dataSource.enqueueEvent(new FeedEvent<>(/* pass as content whatever you need to handle the event, you might need to wrap it in ITimestamped interface */));
}
```

- Then implement a method called `receive(FeedEvent event)`, which gets called by the event loop once the event you inserted in the previous step comes up from the queue. Here, you should send values to outputs and then propagate.

```
        @Override
	public void receive(FeedEvent event) {
		if (event.content instanceof /* whatever you inserted */) {
			sendOutput(event.content);
			getPropagator().propagate();
		} else {
			super.receive(event);
		}
	}
```

isentropy: OK, I'm familiar with that pattern from SendEthereumTransaction. I will update

isentropy:
Not sure why CI is failing:
| Failure:  subscribe starts the polling at correct intervals(com.unifina.feed.PollingMessageSourceSpec)
|  Condition not satisfied after 2.00 seconds and 21 attempts
	at spock.util.concurrent.PollingConditions.within(PollingConditions.java:164)
	at com.unifina.feed.PollingMessageSourceSpec.subscribe starts the polling at correct 

It works locally when I run **grails test-app -unit --stacktrace**

isentropy:
OK CI magically works after I added the comment commit. Must have been a timing issue.

isentropy:
I modified GetEvents to use the enqueueEvent() and receive() pattern above, but it still drops events that are received in quick succession. I'm not sure why. I can see that **receive() is called twice**, but sometimes only 1 displays. When I run in debugger with breakpoint at receive(), both display.

isentropy:
I think this is PR ready for merge, but I have some caveats:

- By calculating nonce using transaction count from PENDING block instead of LATEST, we permit multiple transactions to be submitted before previous transactions are mined. But if a transaction is submitted but ultimately not mined, we will skip over a nonce value and getTransactionCount() won't return the correct nonce. Failed transactions are OK. I think only unmined transactions would cause this issue. @jtakalai raised this point yesterday.

- Transactions that have the same ms timestamp as another might not be displayed. This is a known issue that will be fixed by newer client lib (@hpihkala ).

- I think the table (VariadicEventTable) always uses getGlobals().time.getTime()  and not timestamp from ITimestamped. Is that by design?

isentropy:
I changed SendEthereumTransaction to use block time for transactions too, and that broke test and CI. I will fix now.

hpihkala:
Quick comments (didn't review code changes yet)

> By calculating nonce using transaction count from PENDING block instead of LATEST, we permit multiple transactions to be submitted before previous transactions are mined. But if a transaction is submitted but ultimately not mined, we will skip over a nonce value and getTransactionCount() won't return the correct nonce. Failed transactions are OK. I think only unmined transactions would cause this issue. @jtakalai raised this point yesterday.

Is there a solution to this corner case? @jtakalai 

> Transactions that have the same ms timestamp as another might not be displayed. This is a known issue that will be fixed by newer client lib (@hpihkala ). Note that ethereum block timestamps are in seconds not ms.

Safe to ignore. Please start using the new frontend during development as soon as it gets easier. (It's still missing from `streamr-docker-dev` so it's currently hard).

> I think the table (VariadicEventTable) always uses getGlobals().time.getTime() and not timestamp from ITimestamped. Is that by design?

Yes. In principle modules should always use `getGlobals().time.getTime()` to get the current time while processing an event. In most cases it's the same thing as the timestamp in the event.

> FYI I changed GetEvents and SendEthereumTransaction to use block time for transaction as the ITimestamped timestamp.

First impression: sounds ok, will review and raise concerns if any.

isentropy:
@hpihkala the nonce MUST increase by 1 or TX is rejected. So I think that fear about unmined TX is unfounded. Can I merge?

hpihkala:
@isentropy go ahead!

PR #606, opened 2019-05-22
CORE-1718: initialValue of null converts to object

kare:
- [x] Fix CORE-1718
- [x] Fix: Add missing return statement


hpihkala:
I don’t have bandwidth during the next few days, best to work on this with @harbu

kare:
@hpihkala issue here seems to be Grails inconsistency in JSON parsing:
```
	void "convert json 1"() {
		setup:
		String json = """{"name":null}"""
		when:
		def result = JSON.parse(json)

		String s = new JsonBuilder(result).toString()
		then:
		s == json
	}

	void "convert json 2"() {
		setup:
		String json = """{"name":null}"""
		when:
		def result = new JsonSlurper().parseText(json)

		String s = new JsonBuilder(result).toString()
		then:
		s == json
	}
```

kare:
@hpihkala It also parses invalid JSON to valid Groovy objects:
https://github.com/streamr-dev/engine-and-editor/blob/master/test/unit/com/unifina/controller/api/CanvasApiControllerSpec.groovy#L58

```
-                       json: '{name: "not mine", modules: []}',
+                       json: '{"name": "not mine", "modules": []}',
``` 
https://github.com/streamr-dev/engine-and-editor/blob/master/grails-app/domain/com/unifina/domain/signalpath/Canvas.groovy#L81

hpihkala:
@kare in newer code, I think we're mostly using `Gson` - would it be preferable here too?

PR #609, opened 2019-05-31
distinct eth addresses

mthambipillai:


hpihkala:
This fix alone won't work with the existing data, because it still makes an assumption that the user is unique for a given `id_in_service`. The database doesn't actually currently enforce that, nor does it enforce uniqueness for the `id_in_service` within a `service`... So this is currently possible:

```
user_id | service     | id_in_service | 
1       | ETHEREUM    | 0xF00
1       | ETHEREUM    | 0xF00
1       | ETHEREUM_ID | 0xF00
2       | ETHEREUM    | 0xF00
2       | ETHEREUM_ID | 0xF00
```

We should probably clean up the DB and add constraints so that the address would be unique in the DB.

grails-app/migrations/core/2019-04-01-add-inbox-streams.groovy
```diff
@@ -14,21 +14,23 @@ databaseChangeLog = {
...
+						String userId = userRow['user_id']
+						sql.execute('INSERT INTO permission (version, operation, stream_id, anonymous, user_id) VALUES (0, "read", ?, b\'0\', ?)',
+							ethereumAddress, userId
+						)
+						sql.execute('INSERT INTO permission (version, operation, stream_id, anonymous, user_id) VALUES (0, "write", ?, b\'0\', ?)',
+							ethereumAddress, userId
+						)
+						sql.execute('INSERT INTO permission (version, operation, stream_id, anonymous, user_id) VALUES (0, "share", ?, b\'0\', ?)',
+							ethereumAddress, userId
+						)eth
```
hpihkala:
```suggestion
						)
```

grails-app/domain/com/unifina/domain/security/IntegrationKey.groovy
```diff
@@ -26,6 +26,7 @@ class IntegrationKey implements Serializable {
 	}
 
 	static constraints = {
+		idInService unique: true
```
hpihkala: - Requires new migrations to be generated with the `db_diff` util, so that this will update the database. For that migration to succeed, the database must be cleaned of such occurrences.
- If we add this constraint, then the migration changes are actually unnecessary because duplicate `id_in_service` will be prevented on database level. Of course, the migration can remain as it is now, as the current migration will work in both cases (allow duplicates vs. demand unique).

Adding this constraint is probably the way to go. To apply this change, manual cleaning of the production database is required, and then a new dump from prod to staging must happen to make this work in staging.

hpihkala:
Looks good. Let's merge after staging DB has been updated.

PR #610, opened 2019-05-31
Check for cyclic canvas dependencies

hpihkala:
Cyclic canvas dependencies means that a canvas is a subcanvas of itself. This includes direct dependencies (A is a subcanvas of A), or higher-order cycles (A is a subcanvas of B, which is a subcanvas of A).

Previously in this case, a stack overflow error would occur, as the canvas construction will end up in an infinite loop. Such canvases could not be loaded. This PR introduces a check, and the error gets reported in the canvas json in the usual `moduleErrors` variable, using which it can be communicated to the user.

Also introducing a `moduleError` of type `genericError` which is a generic error not specific to a particular module type (as opposed to for example `compilerError` which is specific to code modules and contains additional information about which line the error occurred on).

```
{
    ...
    "moduleErrors": [
        {
            "type": "genericError",
            "module": -569228597,
            "message": "Cyclic canvas dependency detected!"
        }
    ]
}
```

Also tested manually that the user who reported this issue will be able to open their canvas after this change.


test/unit/com/unifina/signalpath/SignalPathSpec.groovy
```diff
@@ -54,4 +57,32 @@ class SignalPathSpec extends BeanMockingSpecification {
...
+		    	[name: "canvas", value: "root"]
+			]
+		])
+
+		then:
+		ModuleException ex = thrown()
+
+		ex.getModuleExceptions()[0] instanceof CyclicCanvasModuleExceptionMessage
+		ex.getModuleExceptions()[0].getModuleId() == 123
+	}
```
timoxley: Since the code should work for arbitrary levels of nesting maybe should add a test for when recursion is more than one level deep?

kare: @timoxley @hpihkala I tried writing a test like that, but it fails when calling configure() on the first submodule.

timoxley: @kare does that mean there's a bug?

PR #619, opened 2019-06-11
use findAll to avoid NonUniqueResultException

mthambipillai:


hpihkala:
With the uniqueness constraint now in effect for `IntegrationKey.idInService`, how is it actually possible that an `address` would map to multiple `SecUser`s? A particular address should only be in the table exactly once.

PR #620, opened 2019-06-12
CORE-1747 remove Feed layers, refactor event pipeline

hpihkala:
This is a monster PR which has been WIP for a while. It removes more code than it adds by throwing away a lot of unused code and simplifying the data pipeline. This PR also prepares the architecture for [CORE-1624](https://streamr.atlassian.net/browse/CORE-1624), which in turn is a necessary change to make the `engine-and-editor` work with the upcoming P2P Network.

A summary of the changes is as follows:

- Remove the `Feed` layer and all the unnecessary abstraction layers that go with it, such as `MessageSource`, `FieldDetector`, `DataRangeProvider`, `KeyProvider`, and `StreamListener`
- Refactor much of the `DataSource` and `DataSourceEventQueue` (and their subclasses) to be cleaner and simpler
- Refactor how modules can do async event handling via the queue (e.g. HTTP and Ethereum modules). This became much simpler with less requirements on what can be added to the queue (content no longer needs to subclass `ITimestamped`) and more readable with Java 8 lambda `Consumer`s now valid event recipients.
- Make both realtime and historical modes "push" events to the event queue. Previously, realtime was "pushing" and historical was "pulling".
- Remove unnecessary `GlobalsFactory`
- Unify `RealtimeEventQueueMetrics` and `HistoricalEventQueueMetrics` to just `EventQueueMetrics`
- Remove unnecessary interfaces `IEventRecipient`, `IFeedRequirement`, `IStreamRequirement`
- Other minor cleanup

Due to the scope of the changes this should be regarded as a higher-risk PR. It has been manually tested in both realtime and historical mode, but there can always be some issues. It should be further manually tested by reviewers.

**NOTE: the only breaking change for frontend:**

@timoxley The module id of the stream module is no longer rendered in the stream json objects (previously `stream.feed.module`. This is probably used in the frontend to add streams to canvases. For all streams the value is nowadays the same: `147`. This id could be in some global static config as `streamModuleId` or similar. I recommend to grep for `feed.module` in the frontend code and fix all occurences.

**NOTE: JVM crash (EDIT: fixed)**

On my machine, when running via `grails run-app`, running canvases in historical mode many times eventually crashes the JVM with a seg fault. Based on the dumps this has something to do with the hot reloading functionality provided by Spring `ReloadableType`. I suspect this is a symptom that only occurs when hot reloading is active. For example, when compiling the app into a `war` and running it from there (easiest via `grails run-war`), I was not able to reproduce the crash.

I have java `build 1.8.0_161-b12`. Please let me know if you encounter similar problems.


kare:
Something odd is happening here. I've never had a JVM crash while running unit tests:
```
% grails test-app -unit --stacktrace
.
.
.
2019-06-13 14:05:10,718 [Thread-41] INFO  data.RealtimeEventQueue  - runEventLoopUntilDone: starting!
2019-06-13 14:05:11,751 [Thread-41] INFO  data.RealtimeEventQueue  - runEventLoopUntilDone: aborted.
| Running 398 unit tests... 1221 of 1221
2019-06-13 14:05:11,823 [Thread-43] INFO  data.RealtimeEventQueue  - runEventLoopUntilDone: starting!
2019-06-13 14:05:12,839 [Thread-43] INFO  data.RealtimeEventQueue  - runEventLoopUntilDone: aborted.
| Running 398 unit tests... 1222 of 1222
2019-06-13 14:05:12,911 [main] INFO  signalpath.Propagator  - Initialized Propagator with 0 modules in 0 ms.
| Running 398 unit tests... 1223 of 1223
2019-06-13 14:05:12,946 [main] INFO  signalpath.Propagator  - Initialized Propagator with 0 modules in 0 ms.
| Running 398 unit tests... 1224 of 1224
2019-06-13 14:05:12,972 [main] INFO  signalpath.Propagator  - Initialized Propagator with 0 modules in 0 ms.
| Running 398 unit tests... 1225 of 1225
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x0000000101f5012e, pid=33019, tid=0x0000000000003a03
#
# JRE version: Java(TM) SE Runtime Environment (8.0_171-b11) (build 1.8.0_171-b11)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.171-b11 mixed mode bsd-amd64 compressed oops)
# Problematic frame:
# V  [libjvm.dylib+0x15012e]
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /Users/kkn/src/github.com/streamr-dev/engine-and-editor/hs_err_pid33019.log
#
# Compiler replay data is saved as:
# /Users/kkn/src/github.com/streamr-dev/engine-and-editor/replay_pid33019.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
#

```

kare:
I'm guessing BoundedPriorityBlockingQueue.offer() is the culprit
```
% grep BoundedPriorityBlockingQueue hs_err_pid*.log
hs_err_pid33019.log:C1: 121643 26344   !   3       com.unifina.utils.BoundedPriorityBlockingQueue::offer (382 bytes)
hs_err_pid33019.log:Event: 121,624 Thread 0x00007fa535829800 26344   !   3       com.unifina.utils.BoundedPriorityBlockingQueue::offer (382 bytes)
hs_err_pid33340.log:C1:  91561 22765   !   3       com.unifina.utils.BoundedPriorityBlockingQueue::offer (382 bytes)
hs_err_pid33340.log:Event: 91,545 Thread 0x00007faab91b3800 22765   !   3       com.unifina.utils.BoundedPriorityBlockingQueue::offer (382 bytes)
hs_err_pid33793.log:C1:  93529 22512   !   3       com.unifina.utils.BoundedPriorityBlockingQueue::offer (382 bytes)
hs_err_pid33793.log:Event: 93,516 Thread 0x00007fa055005800 22512   !   3       com.unifina.utils.BoundedPriorityBlockingQueue::offer (382 bytes)
hs_err_pid33975.log:C1:  89162 23011   !   3       com.unifina.utils.BoundedPriorityBlockingQueue::offer (382 bytes)
hs_err_pid33975.log:Event: 89,145 Thread 0x00007f8edc009000 23011   !   3       com.unifina.utils.BoundedPriorityBlockingQueue::offer (382 bytes)
```

hpihkala:
Yeah seems  related to what I wrote in the description. Could try without the BoundedPriorityBlockingQueue. It was the only implementation I could find of what I needed, but I have a potential self-rolled workaround in mind if it seems to cause problems.

src/java/com/unifina/data/EventQueueMetrics.java
```diff
@@ -1,5 +1,59 @@
...
+		if (totalEvents == 0) {
+			return 0;
+		} else {
+			return ((double) totalDelay / 1000) / (double) totalEvents;
+		}
+	}
+
+	@Override
+	public String toString() {
+		return String.format("[EventQueueMetrics: events: %s, eventsPerSecond: %s, meanProcessingDelay: %s", numOfEvents, eventsPerSecond(), meanProcessingDelay());
```
kare:
```suggestion
		return String.format("[EventQueueMetrics: events: %d, eventsPerSecond: %.1f, meanProcessingDelay: %.2f", numOfEvents, eventsPerSecond(), meanProcessingDelay());
```

timoxley:
> The module id of the stream module is no longer rendered in the stream json objects (previously stream.feed.module.  

Thanks for the heads up, no changes needed in frontend, we already hard-code the stream id.

src/java/com/unifina/data/RealtimeEventQueue.java
```diff
@@ -3,56 +3,81 @@
...
 
-			// Report processing
 			if (LOGGING_INTERVAL > 0) {
 				elapsedTime += now - startTime;
 				eventCounter++;
 
 				if (eventCounter >= LOGGING_INTERVAL) {
-					double perEvent = (elapsedTime / eventCounter) / 1000.0;
+					double perEvent = (elapsedTime / (double) eventCounter) / 1000.0;
 					log.info("Processed " + eventCounter + " events in " + elapsedTime + " nanoseconds. " +
```
kare:
```suggestion
					log.info(String.format("Processed %d events in %d nanoseconds. That's %s microseconds per event.", eventCounter, elapsedTime, perEvent));
```

src/java/com/unifina/data/RealtimeEventQueue.java
```diff
@@ -3,56 +3,81 @@
...
 			 				 "That's " + perEvent + " microseconds per event.");
```
kare:
```suggestion

```

src/java/com/unifina/data/RealtimeEventQueue.java
```diff
@@ -3,56 +3,81 @@
...
 					eventCounter = 0;
```
kare: eventCounter and elapsedTime are assigned but never used after that

src/java/com/unifina/data/HistoricalEventQueue.java
```diff
@@ -2,188 +2,134 @@
...
-				enqueue(next);
-			}
+			waitIfNecessary(event);
+			handleEvent(event);
 		}
 
 		// Report statistics
-		long feedElapsedTime = System.currentTimeMillis() - feedStartTime;
+		long feedElapsedTime = System.currentTimeMillis() - realTimeStart;
 		log.debug("PERFORMANCE: Processed "+ eventCounter +" events.");
```
kare:
```suggestion
		log.debug(String.format("PERFORMANCE: Processed %d events.", eventCounter));
```

src/java/com/unifina/data/HistoricalEventQueue.java
```diff
@@ -2,188 +2,134 @@
...
-		if (eventCounter >0) {
+		if (eventCounter > 0) {
 			log.debug("PERFORMANCE: Processing took "+((timeSpentProcessing/ eventCounter)/1000.0)+" microseconds per event.");
```
kare: Note missing cast!
```suggestion
			log.debug(String.format("PERFORMANCE: Processing took %s microseconds per event.", (timeSpentProcessing / (double) eventCounter) / 1000.0));```

File: src/java/com/unifina/data/HistoricalEventQueue.java [hunk 7]
```diff
@@ -2,188 +2,134 @@
...
 			log.debug("PERFORMANCE: Entire processing took "+feedElapsedTime+" milliseconds or "+((feedElapsedTime*1000)/ eventCounter)+" microseconds per event.");
```
kare:
```suggestion
			log.debug(String.format("PERFORMANCE: Entire processing took %d milliseconds or %d microseconds per event.", feedElapsedTime, (feedElapsedTime * 1000) / eventCounter));```

File: src/java/com/unifina/data/ClockTick.java [hunk 8]
```diff
@@ -0,0 +1,22 @@
...
+
+	private final Date timestamp;
+
+	public ClockTick(Date timestamp) {
+		this.timestamp = timestamp;
+	}
+
+	@Override
+	public Date getTimestampAsDate() {
+		return timestamp;
```
kare: Just realised that there is a similar problem with the constructor. Real fix would be to store timestamp as a long.
```
package com.unifina.data

import spock.lang.Specification

class ClockTickSpec extends Specification {
	void "does not expose internal representation"() {
		when:
		Date date = new Date(0)
		ClockTick tick = new ClockTick(date)
		tick.getTimestampAsDate().setTime(1)
		then:
		tick.getTimestampAsDate().getTime() == 0
	}
}
```
```suggestion
		return new Date(timestamp.getTime());
```

src/java/com/unifina/data/Event.java
```diff
@@ -0,0 +1,52 @@
+package com.unifina.data;
+
+import java.util.Date;
+import java.util.function.Consumer;
+
+public class Event<ContentClass> implements Comparable<Event<ContentClass>> {
```
kare: What's the definition of ContentClass? It would be easier to read something like:
`public class Event<ContentClass extends SomeOtherClass>`

src/java/com/unifina/feed/FieldDetector.java
```diff
@@ -37,33 +58,59 @@ public void setFlattenMap(boolean flattenMap) {
 			map = new HashMap();
```
kare: Variable map is defined above without the types it holds:
Should read: Map<String, Object> map;

src/java/com/unifina/feed/FieldDetector.java
```diff
@@ -37,33 +58,59 @@ public void setFlattenMap(boolean flattenMap) {
...
 		}
 
-		if (flattenMap) {
+		if (flatten) {
 			map = MapTraversal.flatten(map);
 		}
 
-		List<Map<String, String>> fields = new ArrayList<>();
+		List<FieldConfig> fields = new ArrayList<>();
 		for (Object key : map.keySet()) {
```
kare: It is faster to iterate thought entrySet() than keySet(). This avoids map lookup overhead.
```
		for (Map.Entry<String, Object> entry : map.entrySet()) {
			fields.add(new FieldConfig(entry.getKey(), detectType(entry.getValue())));
```
Not that it makes a huge difference here.

src/java/com/unifina/feed/StreamEventRecipient.java
```diff
@@ -1,25 +1,22 @@
...
+import com.streamr.client.protocol.message_layer.StreamMessage;
+import com.streamr.client.utils.StreamPartition;
 import com.unifina.domain.data.Stream;
+import com.unifina.signalpath.AbstractStreamSourceModule;
 import com.unifina.utils.Globals;
 import com.streamr.client.protocol.message_layer.ITimestamped;
+import org.hibernate.id.Configurable;
 
+import java.util.Collection;
 import java.util.Set;
```
kare: Unused imports

src/java/com/unifina/feed/map/MapMessageEventRecipient.java
```diff
@@ -20,25 +19,24 @@
...
 	}
 
 	@Override
-	protected void sendOutputFromModules(FeedEvent<StreamMessage, ? extends IEventRecipient> event) {
+	protected void sendOutputFromModules(StreamMessage streamMessage) {
 		if (outputsByName == null) {
 			initCacheMap();
 		}
 
 		Map msg;
```
kare: Map<String, Object> msg;

src/java/com/unifina/feed/util/StreamMessageComparator.java
```diff
@@ -0,0 +1,13 @@
+package com.unifina.feed.util;
+
+import com.streamr.client.protocol.message_layer.StreamMessage;
+
+import java.util.Comparator;
+
+public class StreamMessageComparator implements Comparator<StreamMessage> {
```
kare: This class could implement Serializable and to allow it to work with an ordered collection. Ordered collection can be serialized, but only if the comparator is also serializable. Comparator holds no state.

src/java/com/unifina/serialization/SerializationRequest.java
```diff
@@ -0,0 +1,32 @@
...
+import com.unifina.signalpath.SignalPath;
+import grails.util.Holders;
+
+import java.util.Date;
+
+public class SerializationRequest implements ITimestamped {
+
+	private Date timestamp;
+
+	public SerializationRequest(Date timestamp) {
```
kare: Could be private

src/java/com/unifina/data/Event.java
hpihkala: `Event`s can carry arbitrary objects as payloads. I could write

`public class Event<ContentClass extends Object>`

but IMO it would not make the code easier to read.

However I added comments to explain the purpose of `ContentClass` generic, which is to ensure the `content` and `consumer` are compatible.

hpihkala:
> > The module id of the stream module is no longer rendered in the stream json objects (previously stream.feed.module.
> 
> Thanks for the heads up, no changes needed in frontend, we already hard-code the stream id.

@timoxley Note that these unit test mocks won't be accurate anymore, as the `feed` entry will be removed. The mocks should probably be updated to verify that nothing breaks: https://github.com/streamr-dev/streamr-platform/search?q=feed&unscoped_q=feed

hpihkala:
@kare I replaced the `BoundedPriorityBlockingQueue` implementation with a [minimal self-made one](https://github.com/streamr-dev/engine-and-editor/blob/d3408b5c693ea6ba911c0bc6da87037afdec00ba/src/java/com/unifina/utils/BoundedPriorityBlockingQueue.java). After the change I could not reproduce the JVM crash anymore - not in unit tests and not while testing by hand. Looks like it's fixed now.

hpihkala:
Tests passing locally but not on Travis. Will need to debug more...

kare:
@hpihkala 
- No crashes with "grails run-war".
- I'm having trouble stopping running canvases. First click doesn't shutdown the canvas on UI (it is killed on the server). Second click reports it as unreachable. I'm watching server logs here.
- Data and clock (timestamps) seem to be flowing correctly on both historical and realtime modes.
- In historical mode, speed multipliers seem to be working. Verified via visual inspection.

hpihkala:
> I'm having trouble stopping running canvases. First click doesn't shutdown the canvas on UI (it is killed on the server). Second click reports it as unreachable. I'm watching server logs here.

@kare Old or new UI?

src/groovy/com/unifina/signalpath/SignalPathParameter.groovy
```diff
@@ -37,6 +37,7 @@ class SignalPathParameter extends StringParameter {
 			order("lastUpdated", "desc")
 		}
 
+		// TODO: check whether this is till needed in new frontend app
```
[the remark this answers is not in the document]
hpihkala: Checked now: Still needed. Created CORE and PLATFORM tickets (trivial priority) about it and updated this comment.

src/java/com/unifina/data/Event.java
```diff
@@ -0,0 +1,57 @@
...
+		return (t != 0) ? t : Long.compare(sequenceNumber, e.sequenceNumber);
+	}
+
+	@Override
+	public String toString() {
+		return timestamp + " - content: " + content;
+	}
+
+	public void dispatch() {
+		if (consumer != null) {
```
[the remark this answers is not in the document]
hpihkala: Added comment

src/java/com/unifina/data/HistoricalEventQueue.java
```diff
@@ -2,188 +2,126 @@
...
+	private void handleEvent(Event event) {
+		final long time = event.getTimestamp().getTime();
+		final long startTime = System.nanoTime();
 
-		// Never go backwards in time
-		// TODO: this shouldn't be a concern...
+		// Events across different streams/producers aren't necessarily ordered in time.
+		// Never report out-of-order time to modules to prevent weird effects.
+		// Instead, always use the latest observed time as globals.time.
 		if (globals.time==null || time > globals.time.getTime()) {
```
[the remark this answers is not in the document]
hpihkala: I'm not sure, depends on how/if `globals.time` gets initialized elsewhere. It's outside the scope of concerns of this class, so maybe best to leave some defensive code in place?

kare:
@hpihkala new

src/java/com/unifina/data/RealtimeEventQueue.java
```diff
@@ -3,123 +3,72 @@
...
+			} catch (InterruptedException e) {
+				log.error(e);
+			}
+
 			if (event == null) {
-				log.info("doStart: aborting");
-				return;
+				// Timed out while waiting for new events. Just keep trying until aborted.
+				continue;
 			}
```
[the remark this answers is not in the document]
hpihkala: I spent some time on this and cleaned it up. It actually lead to `RealtimeEventQueue` becoming empty/trivial. So I deleted it and instead of three classes I ended up with just two:
- `DataSourceEventQueue`, used by `RealtimeDataSource`
- `HistoricalEventQueue extends DataSourceEventQueue`, used by `HistoricalDataSource`

src/java/com/unifina/data/Event.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.util.Date;
+import java.util.function.Consumer;
+
+/**
+ * An Event is a wrapper class used for async dispatch of content to an attached Consumer.
+ * Event have a natural ordering based on the (timestamp, sequenceNumber) pair.
+ * @param <ContentClass> This is the type of the content, which is arbitrary. The generic is there to ensure that the content and Consumer are compatible.
+ */
+public class Event<ContentClass> implements Comparable<Event<ContentClass>> {
```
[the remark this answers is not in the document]
hpihkala: The use of generics enforces the compatibility of constructor arguments `content` and `consumer`:

```
public Event(ContentClass content, Date timestamp, Consumer<ContentClass> consumer) {
```

This e.g. allows compile-time type checking in `Event` construction statements:

```
new Event<>(new MagnificentItem(), new Date(0), (content) -> {
    // We have compile-time knowledge that content is a MagnificentItem
    doSomethingWithMagnificentItem(content);
});
```

instead of (without generics):

```
new Event(new MagnificentItem(), new Date(0), (content) -> {
    // All we know is that content is an Object
    doSomethingWithMagnificentItem((MagnificentItem) content); // unsafe cast from compiler point of view
});
```

src/java/com/unifina/datasource/DataSource.java
```diff
@@ -1,46 +1,62 @@
...
+public abstract class DataSource implements Consumer<Event> {
 	private static final Logger log = Logger.getLogger(DataSource.class);
 
 	private final Set<SignalPath> signalPaths = new HashSet<>();
 	private final List<IStartListener> startListeners = new ArrayList<>();
 	private final List<IStopListener> stopListeners = new ArrayList<>();
-	private final Map<Long, AbstractFeed> feedById = new HashMap<>();
-	private final Globals globals;
-	private final boolean isHistoricalFeed;
+	protected final Globals globals;
```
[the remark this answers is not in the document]
hpihkala: - `DataSource` and `Globals` have a 1-to-1 relationship
- Multiple `SignalPath`s can run in a `SignalPathRunner` under the same `(DataSource, Globals)` pair (perhaps this is what you're remembering?)

Nowadays we're always running just one `SignalPath` per `SignalPathRunner`. It doesn't add almost any code complexity to support several, but I think I'll remove the possibility anyway since we're not using the feature.

src/java/com/unifina/datasource/DataSource.java
```diff
@@ -1,46 +1,62 @@
...
-	 * Connects a SignalPath to this DataSource. This means connecting to
-	 * all the feeds required by the Modules in the SignalPath and registering
-	 * the Modules with the associated feeds.
-	 * @param sp
+	 * Connects a SignalPath to this DataSource. This has the effect of subscribing the
+	 * DataSource to all the streams required by the Modules in the SignalPath, as well
+	 * as wiring time listeners to the clock.
 	 */
 	public void connectSignalPath(SignalPath sp) {
 		signalPaths.add(sp);
```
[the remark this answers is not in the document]
hpihkala: New `SignalPath`s may be connected to the `DataSource` on the fly. There is just one top-level `SignalPath`, and that's known already before the `DataSource` is started, but for example sub-SignalPaths within a `ForEach` module will be dynamically connected when they are created.

src/java/com/unifina/datasource/DataSource.java
```diff
@@ -1,46 +1,62 @@
...
- * or implicitly according to a SignalPath's needs (via DataSource#connectSignalPath(signalPath)).
+ * DataSource wires together SignalPaths and streams of messages and other Events.
+ * There are two types: HistoricalDataSource and RealtimeDataSource.
  *
- * @author Henri
+ * The DataSource is started with DataSource#start(), which starts the event loop
+ * and blocks until done.
  */
-public abstract class DataSource {
+public abstract class DataSource implements Consumer<Event> {
```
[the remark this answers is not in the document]
hpihkala: Looks like not, it must have been some early idea which didn't stick. I'll remove it.

src/java/com/unifina/datasource/DataSource.java
```diff
@@ -83,69 +98,103 @@ public void addStopListener(IStopListener stopListener) {
...
+								)));
+					}
+
+					@Override
+					public void done() {
+						// Enqueue an end event, which when processed, aborts the event queue.
+						DataSource.this.accept(new Event<>(
+							null,
+							globals.getEndDate() != null ? globals.getEndDate() : new Date(),
+							(nul) -> eventQueue.abort()
```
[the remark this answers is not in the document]
hpihkala: The method is documented in `StreamMessageSource`:

```
	public abstract static class StreamMessageConsumer implements Consumer<StreamMessage> {
		/**
		 * Signals that there will be no further messages from the message source.
		 */
		public abstract void done();
	}
```

I believe it's used in historical mode to signal end of messages. It would indeed be more elegant to call `DataSource#abort()` instead of `eventQueue.abort()`. I'll try it out.

src/java/com/unifina/datasource/DataSourceEventQueue.java
```diff
@@ -50,80 +63,110 @@ public void addDayListener(IDayListener dayListener) {
...
-
-	protected abstract void doStart() throws Exception;
+	/**
+	 * Non-blocking method to asynchronously queue events.
+	 * This is useful for internal event queuing, i.e. cases where
+	 * we want to eventually put something in the event queue even if
+	 * it's congested with incoming data.
+	 */
+	private void enqueueAsync(Event event) {
+		asyncExecutor.submit(() -> {
```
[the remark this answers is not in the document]
hpihkala: It is only used in the special case that the producing thread (the thread in which `DataSourceEventQueue#enqueue()` is called) is the same as the consuming thread. This happens when handling an event causes new events to be queued.

It should never block the thread, because if it does, no events are consumed from the queue, and the whole process is deadlocked.

I changed it to a bounded queue, and added error logging in case it is full. When full, the events-to-be-queued will be dropped, which is the same behavior as with the main (sync) queue if the `offer` times out.

src/java/com/unifina/feed/MessageRouter.java
```diff
@@ -0,0 +1,53 @@
...
+	private static final Logger log = Logger.getLogger(MessageRouter.class);
+	private final Map<StreamPartition, List<Consumer<StreamMessage>>> consumersByStreamPartition = new HashMap<>();
+
+	public Collection<Consumer<StreamMessage>> route(StreamMessage m) {
+		StreamPartition streamPartition = new StreamPartition(m.getStreamId(), m.getStreamPartition());
+		List<Consumer<StreamMessage>> consumersForMessage = consumersByStreamPartition.get(streamPartition);
+		if (consumersForMessage == null) {
+			return Collections.emptyList();
+		} else {
+			return consumersForMessage;
```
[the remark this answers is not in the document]
hpihkala: Currently all methods in this class are called from the `SignalPathRunner` thread, but I'll change the list to a `CopyOnWriteArrayList` to make it future-proof. This choice avoids synchronization on read/traversal (for speed), and contains the assumption that reads far outweigh writes to this list.

src/java/com/unifina/utils/BoundedPriorityBlockingQueue.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.function.Predicate;
+
+/**
+ * Partial implementation of a bounded priority queue. The implementation extends a PriorityBlockingQueue,
+ * adding blocking behavior with the help of a Semaphore.
+ *
+ * NOTE: only offer(E, long, TimeUnit) and poll(long, TimeUnit) are implemented properly.
+ * DO NOT USE other methods to manipulate the queue!
+ */
+public class BoundedPriorityBlockingQueue<E> extends PriorityBlockingQueue<E> {
```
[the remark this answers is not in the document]
hpihkala: The previous option (which caused segfaults) was not part of a commonly used library, rather something copy-pasted from [the internet](https://github.com/gakesson/ConcurrencyUtils/blob/c5794c5b7c0ada763549cadfbbbd345713ace79a/BoundedPriorityBlockingQueue.java).

Expanded comment.

src/java/com/unifina/signalpath/blockchain/GetEvents.java
```diff
@@ -145,7 +131,15 @@ static void convertAndSend(Output output, Object value) {
 	}
 
 	protected void enqueueEvent(LogsResult lr){
-		getGlobals().getDataSource().enqueueEvent(new FeedEvent<LogsResult, GetEvents>(lr, lr.getTimestampAsDate(), this));
+		getGlobals().getDataSource().accept(
```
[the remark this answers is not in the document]
hpihkala: Used in 5 modules - so maybe, but skipping it for now

hpihkala:
@kare following up on this:

> I'm having trouble stopping running canvases. First click doesn't shutdown the canvas on UI (it is killed on the server). Second click reports it as unreachable. I'm watching server logs here.

- I couldn't reproduce this in realtime mode.
- I *was* able to reproduce an issue with aborting historical canvas runs with slow `speed` settings. It should be fixed in the latest commit.

Could you confirm if the problem you noticed is fixed?

kare:
@hpihkala Works for me! Didn't see any issues this time. I was running with "grails run-war".

hpihkala:
This PR seems ready now. However, let's wait until the upcoming live webinar on Sep 4th is over, just in case this breaks something after all. I'll merge this after that.

PR #621, opened 2019-06-13
CORE-1745: Stream-specific inactivity period for detecting stale/non-stale status

kare:


grails-app/domain/com/unifina/domain/data/Stream.groovy
```diff
@@ -35,7 +36,8 @@ class Stream implements Comparable {
 	Boolean autoConfigure = true
 	// Historical data storage period (days)
 	Integer storageDays = DEFAULT_STORAGE_DAYS
-
+	// inactivityThresholdHours is the inactivity period for a stream in hours
+	Integer inactivityThresholdHours = DEFAULT_INACTIVITY_THRESHOLD_HOURS
```
hpihkala: - Should add to `constraints`:

`inactivityThresholdHours(nullable: false, min: 0)`

- Might trigger changes in `db-diff`, probably not

hpihkala: - Should expose in `toMap()`
- Should update `swagger.json` to reflect

hpihkala: Should allow field to be updated in `StreamApiController#update` (similar to `storageDays`)

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -310,9 +310,24 @@ class StreamService {
 		}
 	}
 
+	private Date addThresholdHours(Date date, int hours) {
```
hpihkala: To encapsulate the logic for the special value `0` and keep the code DRY with regard to the threshold calculation, how about adding a helper function to `Stream.groovy`:

```
boolean isStale(Date now, Date latestDataTimestamp) {
    if (inactivityThresholdHours == 0) {
        return false
    } else {
	Calendar calendar = Calendar.getInstance()
	calendar.setTime(now)
	calendar.add(Calendar.HOUR_OF_DAY, -inactivityThresholdHours)
	Date threshold = calendar.getTime()

        return latestDataTimestamp.before(threshold)
    }
}
```

PR #622, opened 2019-06-14
create streams/${id}/subscribers endpoint

mthambipillai:
This new endpoint will be used by the streamr-client during the key exchange for encryption. Before sending the group key encrypted with the subscriber's RSA public key, the publisher first verifies that the request comes from a valid subscriber by checking the list of subscribers using this endpoint.

This endpoint is very similar to "streams/${id}/publishers".

### DIFF
--- a/test/unit/com/unifina/service/StreamServiceSpec.groovy
+++ b/test/unit/com/unifina/service/StreamServiceSpec.groovy
@@ -449,6 +449,39 @@ class StreamServiceSpec extends Specification {
 		addresses == validAddresses
 	}
 
+	void "getStreamEthereumSubscribers should return Ethereum addresses of users with read permission to the Stream"() {
+		setup:
+		service.permissionService = Mock(PermissionService)
+		SecUser user1 = new SecUser(id: 1, username: "u1").save(failOnError: true, validate: false)
+		IntegrationKey key1 = new IntegrationKey(user: user1, service: IntegrationKey.Service.ETHEREUM_ID,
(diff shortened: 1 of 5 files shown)


test/unit/com/unifina/service/StreamServiceSpec.groovy
```diff
@@ -449,6 +449,39 @@ class StreamServiceSpec extends Specification {
...
+		service.permissionService = Mock(PermissionService)
+		SecUser user1 = new SecUser(id: 1, username: "u1").save(failOnError: true, validate: false)
+		IntegrationKey key1 = new IntegrationKey(user: user1, service: IntegrationKey.Service.ETHEREUM_ID,
+			idInService: "0x9fe1ae3f5efe2a01eca8c2e9d3c11102cf4bea57").save(failOnError: true, validate: false)
+		SecUser user2 = new SecUser(id: 2, username: "u2").save(failOnError: true, validate: false)
+		IntegrationKey key2 = new IntegrationKey(user: user2, service: IntegrationKey.Service.ETHEREUM,
+			idInService: "0x26e1ae3f5efe8a01eca8c2e9d3c32702cf4bead6").save(failOnError: true, validate: false)
+		SecUser user3 = new SecUser(id: 3, username: "u3").save(failOnError: true, validate: false)
+
+		// User with key but no write permission - this key should not be returned by the query
```
hpihkala:
```suggestion
		// User with key but no read permission - this key should not be returned by the query
```

PR #624, opened 2019-06-19
fix null user case

mthambipillai:


grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -330,7 +330,7 @@ class PermissionService {
 	private void grantInboxStreamPermissions(SecUser user, Stream stream, Operation operation,
 											 Subscription subscription, Date endsAt, Permission parent) {
 		List<SecUser> otherUsers = getPermissionsTo(stream, operation)*.user
-		otherUsers.removeIf { it.username == user.username }
+		otherUsers.removeIf { it != null && it.username == user.username }
```
hpihkala: Now we leave the `null` items in the `otherUsers` list, which doesn't sound right. Should the condition for removal be instead: 

`it == null || it.username == user.username` ?

PR #628, opened 2019-07-22
Product validation changes

hpihkala:
- Allow Products to be created with default values (previously, was failing with validation error related to `pricePerSecond` not provided), reported by @tumppi 

Additionally fixes for a few issues I noticed while fixing the above:

- Allow `type: "COMMUNITY"` products to be created via the API
- Include the `type` field when rendering products to JSON
- Relax some validation constraints which are no longer compatible with the new product creation flow

cc @kare who is on holiday


grails-app/domain/com/unifina/domain/marketplace/Product.groovy
```diff
@@ -29,7 +29,7 @@ class Product {
 	// The below fields exist in the domain object for speed & query support, but the ground truth is in the smart contract.
 	String ownerAddress
 	String beneficiaryAddress
-	Long pricePerSecond
+	Long pricePerSecond = 0
```
harbu: Do we need default value here if we already have one in `CreateProductCommand`?

src/groovy/com/unifina/api/UpdateProductCommand.groovy
```diff
@@ -47,10 +47,7 @@ class UpdateProductCommand {
 		product.previewStream = previewStream
 		product.previewConfigJson = previewConfigJson
 
-		if (product.pricePerSecond > 0 && product.state == Product.State.NOT_DEPLOYED) {
-			if (pricePerSecond == 0) {
-				throw new InvalidStateException("Cannot update paid Product (price > 0) to be free (price = 0)")
-			}
+		if (product.state == Product.State.NOT_DEPLOYED) {
```
harbu: Have the implications of this change been taken into consideration? There might have been a reason for not allowing paid products to be made free afterwards.

hpihkala: @harbu Not 100% sure what the original reason for this limitation were, but probably this:

Not having to deal with this particular corner case was a handy shortcut back when we were trying to ship the Marketplace in a hurry.

This is not a valid reason anymore, and I can’t think of any technical reason either right now.

grails-app/domain/com/unifina/domain/marketplace/Product.groovy
hpihkala: @harbu I thought the same and actually tried to refactor it to be more DRY, but ended up rolling back and keeping the duplication, as it was already there, and other approaches were causing problems:

- ”Defaults only in domain object”: In the ProductService, currently the *command object* is being validated instead of the domain object. This approach requires the defaults to be in the command object. I tried to change it so that the command object props are applied to the Product first, and then *the Product* is validated, but ran into some weird GORM validation behavior which I couldn’t solve quickly.

- ”Defaults only in command object”: This would be another valid approach, but would break many tests which are doing just new Product(), and would make the default values harder to find in the code.

It could be refactored later but I decided to limit the scope of this PR to the small change I was making.

src/groovy/com/unifina/api/UpdateProductCommand.groovy
harbu: You could run into problems with existing subscribers. If I remember correctly, the way in which paid subscriptions and permissions are represented in the data model differs from the way in which free subscriptions / permissions are represented.

hpihkala: Can you point out how exactly a problem would occur?

tumppi: What should we do with the contract when paid product is updated to free? Do we just let it hang there or is it removed from the chain? There's also validation logic for pricePerSecond in the contract: https://github.com/streamr-dev/marketplace-contracts/blob/master/contracts/Marketplace.sol#L232

harbu: I searched through Technical Debt tickets, closest thing I could find related to this is CORE-1322, which is worth a read.

 A couple of things spring to mind:
- The streams of free products have associated anonymous READ permissions in the system that are not necessarily present for paid products. These get created when the free product is deployed / published. (And revoked when the free product is undeployed.)
- Subscriptions of paid products are of class `PaidSubscription` and those of free products are `FreeSubscription`. You could potentially run into problems if a `PaidSubscription` is suddenly referring to a free Product.

It's a long time since I touched this code, so I can't really come up with anything more concrete. All I know is that there are notable differences in how free and paid products are handled and the associated domain objects that are created.

I would say it's okay to proceed with this but expect to discover some bugs later on. At this point in time, without further knowledge, it's probably the practical way forward.

tumppi:
Could we get this merged so we could start testing new frontend features with a real backend?

src/groovy/com/unifina/api/UpdateProductCommand.groovy
hpihkala: Proceeding with this to unblock frontend development. As Eric noted, this may lead to some problems in free vs. paid products behaving differently, but let's resolve any issues if/when we hit them.

PR #594, opened 2019-05-02
check DATAcoin balance

isentropy:
API endpoint (/api/v1/users/me/balance) to check summed datacoin balance

https://streamr.atlassian.net/browse/CORE-1187

PR #631, opened 2019-08-05
DATAcoin balance PR changes

isentropy:
This new PR replaces old PR https://github.com/streamr-dev/engine-and-editor/pull/594, which was way out of sync with E&E.


isentropy:
I think this is good to go, but let me just test manually to see that it's still working.

isentropy:
manual test with parity looks fine. Can I merge?

PR #636, opened 2019-08-14
init migration

mthambipillai:
The previous inbox streams migration created read/write/share permissions for every user to its own inbox stream.

What's missing is to give, for every stream, the following permissions:
- every subscriber can write to the inbox stream of every publisher
- every publisher can write to the inbox stream of every subscriber


grails-app/migrations/core/2019-08-14-add-inbox-permissions.groovy
```diff
@@ -0,0 +1,36 @@
...
+databaseChangeLog = {
+	changeSet(author: "mthambipillai", id: "adding-permissions-to-inbox-streams") {
+		grailsChange {
+			change {
+				sql.eachRow('SELECT DISTINCT id_in_service FROM integration_key WHERE service = \"ETHEREUM_ID\" OR service = \"ETHEREUM\"') { row ->
+					String publisherInboxAddress = row['id_in_service']
+					sql.eachRow('SELECT user_id FROM integration_key WHERE id_in_service = :address', [address:publisherInboxAddress]) { userRow ->
+						// for each user with an inbox stream
+						String publisherId = userRow['user_id']
+						sql.eachRow('SELECT stream_id FROM permission WHERE stream_id IS NOT NULL AND user_id = :user_id AND operation = \"write\" AND inbox = b\'0\'', [user_id:publisherId]) { streamRow ->
```
hpihkala: Tests pass, but look like there isn't data in the test database that this migration would actually run against.

I tried running it against a dump of the production db, which revealed this error:

```
     Reason: com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: Unknown column 'inbox' in 'where clause':
          Caused By: Unknown column 'inbox' in 'where clause'
```

(there is no field called `inbox` in the `permission` table`)

hpihkala:
Not relevant anymore after key exchange refactoring, closing

PR #647, opened 2019-08-27
CORE-1762: Fix Stream Module

kare:


grails-app/services/com/unifina/service/SignalPathService.groovy
```diff
@@ -111,8 +113,12 @@ class SignalPathService {
 		// can be problematic when collaborating on shared canvas; though even then it makes sense to force
 		//   explicit read rights sharing to streams on that canvas
 		for (Stream s in sp.getStreams()) {
+			boolean streamModuleHasNoStreamAssociated = s == null
```
hpihkala: Wouldn't it be more elegant to contain this logic in `ConfigurableStreamModule` instead of leaking this module-specific check to this generic part of code?

Basically we want to throw if the `ConfigurableStreamModule` is being created in a run context (realtime or historical) but the stream is `null`. In `ConfigurableStreamModule#onConfiguration()` (after line 42) you could do something like:

```
if (stream == null && getGlobals().isRunContext()) {
    throw new ....
}
```

Maybe that's all it takes? (untested)

kare: @hpihkala `isRunContext()` is false. `ConfigurableStreamModule#onConfiguration()` is executed when StreamModule is dragged and dropped to canvas (without config).

PR #649, opened 2019-08-28
CORE-1782: Fix deleteDataUpTo endpoint

kare:


web-app/misc/swagger.json
```diff
@@ -3066,9 +3066,16 @@
           {
             "in": "query",
```
hpihkala: Looks like the issue was that the frontend is now passing the parameters in the request body (judging from the code changes), not as query params, right? 

The `swagger.json` should reflect this change by changing the `in` field to `body` (if I remember correctly).

web-app/misc/swagger.json
```diff
@@ -3066,9 +3066,16 @@
...
             "name": "date",
-            "description": "Delete data up to date. Format is either milliseconds since epoch or ISO8601 datetime",
+            "description": "Delete data up to date. Format is either milliseconds (integer) since epoch or ISO8601 datetime (string).",
             "required": true,
             "type": "string"
+          },
+          {
+            "in": "query",
```
hpihkala: same

web-app/misc/swagger.json
```diff
@@ -3114,17 +3121,30 @@
           {
             "in": "query",
```
hpihkala: same

web-app/misc/swagger.json
```diff
@@ -3114,17 +3121,30 @@
...
             "name": "start",
-            "description": "Start date where the data deletion starts. Format is either milliseconds since epoch or ISO8601 datetime.",
-
+            "description": "Start date where the data deletion starts. Format is either milliseconds (integer) since epoch or ISO8601 datetime (string).",
             "required": true,
             "type": "string"
           },
+          {
+            "in": "query",
```
hpihkala: same

web-app/misc/swagger.json
```diff
@@ -3114,17 +3121,30 @@
...
+            "name": "start",
+            "description": "Start date where the data deletion starts. Format is either milliseconds (integer) since epoch or ISO8601 datetime (string)",
+            "required": true,
+            "type": "integer"
+          },
           {
             "in": "query",
```
hpihkala: same

web-app/misc/swagger.json
```diff
@@ -3114,17 +3121,30 @@
...
             "name": "end",
-            "description": "Delete data up to this date. Format is either milliseconds since epoch or ISO8601 datetime.",
+            "description": "Delete data up to this date. Format is either milliseconds (integer) since epoch or ISO8601 datetime (string).",
             "required": true,
             "type": "string"
+          },
+          {
+            "in": "query",
```
hpihkala: same

web-app/misc/swagger.json
timoxley: > the frontend is now passing the parameters in the request body

Very curious, @juhah do you recall why this change from query params to request body? https://github.com/streamr-dev/streamr-platform/commit/b52775d1b4#diff-c245c1fe624a2e09cee0cbe7e228fa1aL515 Seems like this never should have worked, maybe you were working off a WIP branch or something in https://github.com/streamr-dev/streamr-platform/pull/408 We should get all the endpoints the frontend is using under test.

After this change, does that mean that if someone hits the endpoint with query params instead then they'll get the `NullPointerException` `500`?

Backend should probably do some actual validation of request data e.g. "date parameter missing from {request body|query parameters}" and responding with a `4xx` error. Grails/Spring likely provides some standardised tools for doing exactly this.

hpihkala: Let’s keep the body params, it’s actually better in line with the other endpoints.

I doubt anyone is calling this endpoint apart from our own UI. Ok to make the breaking (lol, actually fixing) change.

kare:
Swagger is now fixed. Fixed some other validation errors while at it.

kare:
```
% curl -v -X DELETE -H "Content-Type: application/json" -H "Authorization: token tester1-api-key" "http://localhost:8081/streamr-core/api/v1/streams/-VMAuwLtSx25LOv_jchNJg/deleteDataUpTo"
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8081 (#0)
> DELETE /streamr-core/api/v1/streams/-VMAuwLtSx25LOv_jchNJg/deleteDataUpTo HTTP/1.1
> Host: localhost:8081
> User-Agent: curl/7.54.0
> Accept: */*
> Content-Type: application/json
> Authorization: token tester1-api-key
>
< HTTP/1.1 400 Bad Request
< Server: Apache-Coyote/1.1
< Content-Type: application/json;charset=UTF-8
< Transfer-Encoding: chunked
< Date: Fri, 30 Aug 2019 11:19:45 GMT
< Connection: close
<
* Closing connection 0
{"code":"PARAMETER_MISSING","message":"Unable to parse a date from field 'date' with value 'null'"}
%
```

PR #650, opened 2019-08-28
Access realtime/historical data via API instead of directly from Redis/Cassandra

hpihkala:
This PR makes `engine-and-editor` access stream content via the API instead of directly. This is necessary for decoupling the Engine from the Network implementation in preparation for the Corea network release.

- Canvases now subscribe to data via the API instead of accessing Redis/Cassandra directly for realtime/historical runs respectively.
- Canvases now publish data via the API instead of producing to Kafka directly.
- To achieve the above, a shared `StreamrClient` instance is available to the `DataSource` and canvas modules via `Globals`. The `StreamrClient` authenticates as the user who started the canvas.
- Remove all references to Kafka from the codebase.
- `RunCanvasSpec` couldn't work anymore, as the API is not up in integration tests. Ported the same test content to an end-to-end JS test.

**Also fixed some loosely related stuff that I encountered while working on the above, sorry for the noisy PR:**

- [CORE-1781](https://streamr.atlassian.net/browse/CORE-1781): Add partition selection support to `SendToStream` module.
- Throttle the rate at which messages are published during a csv upload.
- Allow stream partition count to be updated via the API.
- `PUT` requests to stream endpoint now behave the same as other PUT endpoints: return `200` and render the updated resource (was: return 204 and empty response).
- Other minor codebase maintenance and README updates.

**Out of scope for this PR:**

- There are still some direct usage of Cassandra in `StreamService` related to deleting messages. These can't be removed yet, as the P2P network API doesn't support deleting messages yet. In the Corea network we'll initially have just one centralized Cassandra cluster for storage, so the existing functionality in `engine-and-editor` will work, and it is not blocking the Corea network from being deployed. The delete support is outlined in [NET-57](https://streamr.atlassian.net/browse/NET-57).

**Notable shortcomings/hacks:**

In order to access data via the API, the connection needs to be authenticated. So the canvas needs to authenticate as the user who started it. For this purpose, the running canvas will need to use some credential of the user to authenticate the session. Currently an arbitrarily selected API key is used (all users have at least one API key for some other, earlier reason, but now this requirement is enforced by this implementation).

This approach may have some problems in various corner cases. Such as when an API key is revoked. If that key happens to be used by a running canvas, and the session token created when the canvas was started happens to expire, then the API key is used again to start a new session. For a revoked API key, this procedure will fail, and the canvas will no longer be able to interact with the API.


hpihkala:
@kare feedback addressed! May I also ask for quite a lot of manual testing of canvases on this one, both realtime and historical please.

kare:
@hpihkala There's some failing specs. Not sure if its relevant?

```
com.unifina.api.CreateProductCommand.getProperties()Ljava/util/Map;
java.lang.NoSuchMethodError: com.unifina.api.CreateProductCommand.getProperties()Ljava/util/Map;
	at com.unifina.service.ProductService.create(ProductService.groovy:127)
	at com.unifina.service.ProductServiceSpec.create() creates and returns Product with correct info and NEW state(ProductServiceSpec.groovy:243)
```

kare:
@hpihkala I did some light testing (realtime/historical) today. Seems to work! I'll continue tomorrow with some heavy testing

Only thing I spotted was this (e&e logs):
`2019-09-17 17:01:10,852 [OkHttp ConnectionPool] WARN  okhttp3.OkHttpClient  - A connection to http://localhost:8081/ was leaked. Did you forget to close a response body? To see where this was allocated, set the OkHttpClient logger level to FINE: Logger.getLogger(OkHttpClient.class.getName()).setLevel(Level.FINE);`
Didn't @mthambipillai just fix something similar?

hpihkala:
@kare The tests are passing on Travis - how did you get this error? Could it be a local problem, try `grails clean`?

> @hpihkala There's some failing specs. Not sure if its relevant?
> 
> ```
> com.unifina.api.CreateProductCommand.getProperties()Ljava/util/Map;
> java.lang.NoSuchMethodError: com.unifina.api.CreateProductCommand.getProperties()Ljava/util/Map;
> 	at com.unifina.service.ProductService.create(ProductService.groovy:127)
> 	at com.unifina.service.ProductServiceSpec.create() creates and returns Product with correct info and NEW state(ProductServiceSpec.groovy:243)
> ```

kare:
@hpihkala It was a local problem. Now unit and integration tests pass. I use `scripts/git-change-branch`. It's not a 100% solution and cause for failure is difficult to detect (grails compiler error vs. error in code).

PR #655, opened 2019-09-05
network info passed via Contract object

isentropy:
- added network field to the Contract object
- SolidityCompileDeploy and GetEtherumContractAt both populate Contract.network with network from their EthereumModuleOptions
- SendEthereumTransaction and GetEvents overwrite their EthereumModuleOptions.network with network from Contract

Tested locally and working.

SendEthereumTransaction and GetEvents still prompt for network choice, although it is now ignored. I think that needs to be fixed on frontend  @timoxley 

see https://streamr.atlassian.net/browse/CORE-1787


timoxley:
@isentropy if you're talking about this option in the sidebar, that's defined by the backend.

image

isentropy:
> @isentropy if you're talking about this option in the sidebar, that's defined by the backend.
> 
> image

@jtakalai or @hpihkala , do you know how to remove the network option from the sidebar? I'd like to remove the option from GetEvents and EthereumCall modules only.

hpihkala:
@isentropy Sure, it’s defined, read, and handled in the module code. You should be able to easily find it: probably in onConfiguration() and getConfiguration(). It may be in the superclass, as it was built as a shared thing for all Ethereum modules!

If we now need it only for some of those modules, you’d need to move it down one abstraction level, or alternatively keep it in the superclass but opt out of it in the children that don’t need it.

isentropy:
I removed the network selector option from the EthereumCall and GetEvents modules. Everything is working and I think this is good to go now.

grails-app/conf/Config.groovy
```diff
@@ -269,9 +269,22 @@ environments {
...
+streamr.ethereum.networks = PropertiesUtil.matchingPropertiesToMap("streamr.ethereum.networks.", System.getProperties()) ?: [
+	local: "http://localhost:8545",
+	rinkeby: "https://rinkeby.infura.io/v3/117d24d0fef1434fbd14567435a363d9",
+	mainnet: "https://mainnet.infura.io/v3/117d24d0fef1434fbd14567435a363d9",
+]
+streamr.ethereum.wss = PropertiesUtil.matchingPropertiesToMap("streamr.ethereum.wss.", System.getProperties()) ?: [
+	local: "ws://localhost:8545",
+	rinkeby: "wss://rinkeby.infura.io/ws/v3/117d24d0fef1434fbd14567435a363d9",
+	mainnet: "wss://mainnet.infura.io/ws/v3/117d24d0fef1434fbd14567435a363d9",
+]
```
isentropy: @jtakalai did you mean to check in Infura URLs? Are these public?

jtakalai: Yea but restricted to DATA and Marketplace contract. Feel free to delete. It was for testing the different networks with dev docker setup, but maybe not necessary for tests or anything.

PR #659, opened 2019-09-17
CORE-1797: Add pendingChanges field to Product

kare:


src/groovy/com/unifina/api/UpdateProductCommand.groovy
```diff
@@ -24,13 +27,16 @@ class UpdateProductCommand {
 	Product.Currency priceCurrency
 	Long minimumSubscriptionInSeconds
 
+	PermissionService permissionService
```
hpihkala: Hmm, looks sort of weird to have `permissionService` listed among the content fields. Would it theoretically be possible for an attacker to try and inject something to this field with a payload like:

```
{
  "permissionService": //something
  ...
}
```

If the above is at all a valid concern (I'm not saying it is), the service can also be accessed statically in `updateProduct` via `Holders.getApplicationContext().getBean(StreamrClientService.class)`.

test/unit/com/unifina/api/UpdateProductCommandSpec.groovy
```diff
@@ -181,10 +182,26 @@ class UpdateProductCommandSpec extends Specification {
...
 
 		when:
-		command.updateProduct(product)
+		command.updateProduct(product, new SecUser())
 
 		then:
 		thrown(FieldCannotBeUpdatedException)
 	}
 
+	void "updateProduct() checks share permission when pendingChanges field is given"() {
```
hpihkala: Successful case should be tested too?

src/groovy/com/unifina/api/UpdateProductCommand.groovy
kare: There was some (old) discussion on stackoverflow. I'll pass it as a parameter.

test/unit/com/unifina/api/UpdateProductCommandSpec.groovy
```diff
@@ -195,13 +197,26 @@ class UpdateProductCommandSpec extends Specification {
...
 		when:
-		command.updateProduct(product, new SecUser())
+		command.updateProduct(product, new SecUser(), permissionService)
 		then:
-		1 * command.permissionService.canShare(_, product) >> false
+		1 * permissionService.canShare(_, product) >> false
 		thrown(FieldCannotBeUpdatedException)
 	}
+
+	void "updateProduct() doesn't check sharing permission when pendingChanges is not given"() {
```
hpihkala: This test is good to have too, but in my request for an additional test, I meant the case where:
- `canShare(_, product) >> true`
- check that the `pendingChanges` field was updated

hpihkala: I had this open so went ahead and committed this

PR #665, opened 2019-09-20
Make grails accept {"type": "community"}

jtakalai:
Real fix would be to make it case-insensitive somehow. Or just change the spec.


jtakalai:
maybe just keep ALL_CAPS since front-end has started using that. I'll try to amend all specs.

PR #666, opened 2019-09-23
make method synchronized

mthambipillai:


mthambipillai:
@hpihkala yes. I created 10 threads to make authenticated requests in parallel, 10 initialisations of redis without this fix, only 1 with this fix

hpihkala:
@mthambipillai 👍

PR #670, opened 2019-09-25
fix WS reconnection err

isentropy:
- WS connector applies filter on reconnect and "no filter" error
- RPC errors checked, logged, no exception thrown
- check for null timestamp within non-null block

### DIFF
--- a/src/java/com/unifina/signalpath/blockchain/ContractEventPoller.java
+++ b/src/java/com/unifina/signalpath/blockchain/ContractEventPoller.java
@@ -32,6 +32,7 @@ class ContractEventPoller implements Closeable, Runnable, JsonRpcResponseHandler
 	private final String contractAddress;
 	private final EventsListener listener;
 	private String filterId;
+	//keepPolling starts true, and only turns false when close() is called
 	private boolean keepPolling = true;
 
 
@@ -86,7 +87,7 @@ class ContractEventPoller implements Closeable, Runnable, JsonRpcResponseHandler
 		List params = singletonList(singletonMap("address", contractAddress));
 		try {
 			rpc.rpcCall("eth_newFilter", params, ID_ADDFILTER);
-			log.info("adding new filter");
+			log.info("adding new filter to contract address " + contractAddress);
 		} catch (Exception e) {
 			listener.onError(e.getMessage());
 			throw new RuntimeException(e);
--- a/src/java/com/unifina/signalpath/blockchain/ContractEventPoller.java
+++ b/src/java/com/unifina/signalpath/blockchain/ContractEventPoller.java
@@ -116,15 +116,7 @@ class ContractEventPoller implements Closeable, Runnable, JsonRpcResponseHandler
 		try {
 			log.info(String.format("Polling filter '%s'.", filterId));
 			rpc.rpcCall("eth_getFilterChanges", singletonList(filterId), ID_POLLFILTER);
-		} catch (HttpEthereumJsonRpc.ErrorObjectException e) {
-			if (RPC_CODE_NO_FILTER == e.getCode()) {
-				log.info("Resetting filter...");
-				filterId = null;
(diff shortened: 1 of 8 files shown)


src/java/com/unifina/signalpath/blockchain/ContractEventPoller.java
```diff
@@ -31,10 +32,13 @@
 	private final String contractAddress;
 	private final EventsListener listener;
 	private String filterId;
-	private boolean keepPolling = false;
+	private boolean keepPolling = true;
```
isentropy: keepPolling starts true, and only turns false when close() is called

src/java/com/unifina/signalpath/blockchain/HttpEthereumJsonRpc.java
```diff
@@ -14,6 +14,7 @@
 class HttpEthereumJsonRpc extends EthereumJsonRpc{
 	HttpEthereumJsonRpc(String url, JsonRpcResponseHandler handler) {
 		super(url, handler);
+		handler.init();
```
isentropy: in this instance handler.init() installs filter

src/java/com/unifina/signalpath/blockchain/ContractEventPoller.java
hpihkala: The above comment would be useful as a comment in the code.

src/java/com/unifina/signalpath/blockchain/ContractEventPoller.java
```diff
@@ -79,7 +86,7 @@ private void newFilter() {
 		List params = singletonList(singletonMap("address", contractAddress));
 		try {
 			rpc.rpcCall("eth_newFilter", params, ID_ADDFILTER);
-			keepPolling = true;
+			log.info("adding new filter");
```
hpihkala: If we're gonna log, then let's include the `contractAddress` in the logged message.

src/java/com/unifina/signalpath/blockchain/HttpEthereumJsonRpc.java
hpihkala: Here too it would be better to comment in code instead of here :)

src/java/com/unifina/signalpath/blockchain/Web3jHelper.java
```diff
@@ -201,7 +201,18 @@ public static long getBlockTime(Web3j web3j, TransactionReceipt tr) throws IOExc
 			log.error("Error fetching block " + dbp + ". Error = " + eb.getError());
 			return -1;
 		}
-		long ts = eb.getBlock().getTimestamp().longValue();
+		EthBlock.Block block = eb.getBlock();
+		if(block == null){
+			log.error("getBlock() returned null block for txHash " + tr.getTransactionHash());
+			return -1;
```
hpihkala: Returning an unexpected value to represent an error condition probably isn't the best approach. A careless user of `getBlockTime` could easily overlook to check for a value `-1` and accidentally use that as the timestamp. Better would be to throw a declared exception (e.g. `BlockWasNullException`) to force the calling code to handle the situation appropriately.

src/java/com/unifina/signalpath/blockchain/Web3jHelper.java
```diff
@@ -201,7 +201,18 @@ public static long getBlockTime(Web3j web3j, TransactionReceipt tr) throws IOExc
...
+		}
+		BigInteger timestamp = block.getTimestamp();
+		if(timestamp == null){
+			log.error("getBlock() returned null timestamp for block: "+block+", re txHash: " + tr.getTransactionHash());
+			return -1;
```
hpihkala: Same here, could throw a declared exception `BlockTimestampIsNullException`.

PR #671, opened 2019-09-26
use SecureRandom instead of Random

mthambipillai:


mthambipillai:
@hpihkala the following now use the same utility function implemented with `SecureRandom`:
- `SessionToken` generation
- `Challenge` generation
- initial user password generation

The 5 other generated secrets in your list all rely on `UUID.randomUUID()` which is implemented with `SecureRandom` so there is no insecure PRNG problem there.

src/java/com/unifina/security/Challenge.java
```diff
@@ -1,8 +1,10 @@
 package com.unifina.security;
 
+import com.unifina.utils.AlphanumericStringGenerator;
 import org.apache.commons.lang.RandomStringUtils;
 import org.joda.time.DateTime;
 
+import java.security.SecureRandom;
```
hpihkala: import no longer used

src/java/com/unifina/security/SessionToken.java
```diff
@@ -1,8 +1,10 @@
 package com.unifina.security;
 
+import com.unifina.utils.AlphanumericStringGenerator;
 import org.apache.commons.lang.RandomStringUtils;
 import org.joda.time.DateTime;
 
+import java.security.SecureRandom;
```
hpihkala: import no longer used

PR #675, opened 2019-10-01
CORE-1794-compile-without-account

isentropy:
- Compile button works when no account is selected in SolidityCompileDeploy, So you can see if your code compiles, but not much else. I think a [copy ABI button](https://streamr.atlassian.net/browse/PLATFORM-1023) would be useful here to copy the generated ABI into a GetContractAt module.
- Compile shows errors in this case
- Deploy throws an exception when no account selected. I think it should be **disabled** if no Ethreum account is selected. Why show an option that just throws an error?


isentropy:
@hpihkala added unit tests:

1. compile works with null ethAcct
2. deploy doesn't work with null ethAcct

please tell me if this looks ok

PR #680, opened 2019-10-04
generate private key if none provided

isentropy:
generates ethereum private key if none provided, and exports it in plaintext in EthereumKey.jsonMap(). I tested manually. Still needs a frontend hook to copy the private key.


grails-app/services/com/unifina/service/EthereumIntegrationKeyService.groovy
```diff
@@ -51,6 +63,7 @@ class EthereumIntegrationKeyService {
 				idInService: address,
 				json: ([
 					privateKey: encryptedPrivateKey,
+					privateKeyPlaintext: privateKey,
```
hpihkala: The private key should not be stored in plaintext, so please remove this line

grails-app/domain/com/unifina/domain/security/IntegrationKey.groovy
```diff
@@ -48,7 +48,11 @@ class IntegrationKey implements Serializable {
 	@CompileStatic
 	private Map jsonMap() {
 		if (service == Service.ETHEREUM || service == Service.ETHEREUM_ID) {
-			return [address: ((JSONObject) JSON.parse(json)).get("address")]
+			JSONObject jso = (JSONObject) JSON.parse(json)
+			Map jsmap = [address: jso.get("address")]
+			if(jso.has("privateKeyPlaintext"))
```
hpihkala: If `service` is `ETHEREUM`, we should read and decrypt the encrypted private key from the `privateKey` field, and write it to the resulting map (perhaps as key `privateKey` instead of `privateKeyPlaintext` for brevity)

grails-app/domain/com/unifina/domain/security/IntegrationKey.groovy
```diff
@@ -17,6 +22,13 @@ class IntegrationKey implements Serializable {
 	Date dateCreated
 	Date lastUpdated
 
+	@PostConstruct
+	void init() {
+		String password = grailsApplication.config["streamr"]["encryption"]["password"]
+		Assert.notNull(password, "streamr.encryption.password not set!")
+		encryptor = new StringEncryptor(password)
+	}
+
```
isentropy: @hpihkala What is the correct way to get the decryption password in IntegrationKey.groovy? I tried this idiom, which works in EthereumIntegrationKeyService.groovy to inject the grailsApplication bean, but it doesnt work here, maybe bc IntegrationKey.groovy isn't a service.

grails-app/domain/com/unifina/domain/security/IntegrationKey.groovy
```diff
@@ -48,7 +60,13 @@ class IntegrationKey implements Serializable {
 	@CompileStatic
 	private Map jsonMap() {
 		if (service == Service.ETHEREUM || service == Service.ETHEREUM_ID) {
-			return [address: ((JSONObject) JSON.parse(json)).get("address")]
+			JSONObject jso = (JSONObject) JSON.parse(json)
+			Map jsmap = [address: jso.get("address")]
+			if(service == Service.ETHEREUM) {
+				String decryptedPrivateKey = encryptor.decrypt(jso.getString("privateKey"), user.id.byteValue())
```
isentropy: isnt 1 byte of salt a bit light?

isentropy:
Does this look almost right?  Still not working:
1. how do I get password within IntegrationKey.groovy class?
2. need to fix unit tests. EthereumIntegrationKeyServiceSpec (changes not yet committed) works when I manually set password to "password" in IntegrationKey.groovy, and I need to fixIntegrationKeySpec.groovy

grails-app/domain/com/unifina/domain/security/IntegrationKey.groovy
isentropy: should StringEncryptor be a transient member var or created in jsonMap()?

isentropy: EthereumIntegrationKeyService keyService = Holders.getApplicationContext().getBean(EthereumIntegrationKeyService.class)
keyService.decryptPrivateKey(getValue());
?

isentropy:
I think think this is code complete now @hpihkala , @harbu. 
Changes:
- EthereumIntegrationKeyService creates new key if passed null private key
- IntegrationKey decrypts the encrypted "privateKey" field created by EthereumIntegrationKeyService and outputs a plaintext "privatekey" in jsonMap(). IntegrationKey object stores a transient EthereumIntegrationKeyService for decryption
- changes some bean creation methods
- unit test fixes

grails-app/domain/com/unifina/domain/security/IntegrationKey.groovy
```diff
@@ -50,8 +58,10 @@ class IntegrationKey implements Serializable {
 		if (service == Service.ETHEREUM || service == Service.ETHEREUM_ID) {
 			JSONObject jso = (JSONObject) JSON.parse(json)
 			Map jsmap = [address: jso.get("address")]
-			if(jso.has("privateKeyPlaintext"))
-				jsmap.put("privateKeyPlaintext", jso.get("privateKeyPlaintext"))
+			if(service == Service.ETHEREUM) {
+				String decryptedPrivateKey = getEthereumIntegrationKeyService().decryptPrivateKey(this)
```
hpihkala: To me, feels like caching the the `ethereumIntegrationKeyService` is an unnecessary micro-optimization, probably not worth the slight code pollution. I don't see a problem with:

`String decryptedPrivateKey = Holders.getApplicationContext().getBean(EthereumIntegrationKeyService.class).decryptPrivateKey(this)`

grails-app/services/com/unifina/service/EthereumIntegrationKeyService.groovy
```diff
@@ -32,9 +33,26 @@ class EthereumIntegrationKeyService {
...
+
+	//always instantiates
+	private initStringEncryptor(){
+		encryptor = new StringEncryptor(getPassword())
+	}
+
+	// instantiates only if encryptor doesn't exist
+	private StringEncryptor getStringEncryptor() {
+		if (encryptor == null) {
+			initStringEncryptor()
```
hpihkala: I guess you could remove the one-liner function `initStringEncryptor` by simply replacing this line with:
```
encryptor = new StringEncryptor(getPassword())
```
and calling `getStringEncryptor()` on line 36 instead of `initStringEncryptor()`

grails-app/services/com/unifina/service/EthereumIntegrationKeyService.groovy
```diff
@@ -32,9 +33,26 @@ class EthereumIntegrationKeyService {
 
 	@PostConstruct
 	void init() {
-		String password = grailsApplication.config["streamr"]["encryption"]["password"]
+		initStringEncryptor()
```
hpihkala: I'm actually wondering why the changes in this class were necessary? What was the problem before the changes?

test/unit/com/unifina/controller/api/IntegrationKeyApiControllerSpec.groovy
```diff
@@ -20,6 +20,8 @@ class IntegrationKeyApiControllerSpec extends ControllerSpecification {
 	SecUser someoneElse
 
 	def setup() {
+		EthereumIntegrationKeyService ethereumIntegrationKeyService = new EthereumIntegrationKeyService()
```
hpihkala: - Shadows the class field with the same name `ethereumIntegrationKeyService`
- Should prefer mocks in unit tests instead of real collaborators

Which test case fails without this? 

Looks like in this test suite, the `EthereumIntegrationKeyService` is commonly mocked locally in each test cases, e.g. like in line 123. Not sure why this repetition is there. The mock could also be registered for all tests in `setup` with:

```
ethereumIntegrationKeyService = mockBean(EthereumIntegrationKeyService, Mock(EthereumIntegrationKeyService))
```

One of the tests use `Stub` instead of `Mock`, not sure why. Maybe this is the reason for the repetition (one test case needs a different setup).

test/unit/com/unifina/domain/security/IntegrationKeySpec.groovy
```diff
@@ -1,16 +1,26 @@
...
-class IntegrationKeySpec extends Specification {
+import com.unifina.BeanMockingSpecification
+import com.unifina.service.EthereumIntegrationKeyService
+import grails.test.mixin.Mock
+import grails.test.mixin.TestFor
+@TestFor(IntegrationKey)
+@Mock([SecUser])
+class IntegrationKeySpec extends BeanMockingSpecification {
+	def setup() {
+		EthereumIntegrationKeyService ethereumIntegrationKeyService = new EthereumIntegrationKeyService()
```
hpihkala: Prefer mocks to real collaborators

grails-app/services/com/unifina/service/EthereumIntegrationKeyService.groovy
isentropy: When I revert it to the original, it fails to load the password correctly and decrypt key in IntegrationKeySpec:24

`| Failure:  toMap().json for Ethereum integration key(com.unifina.domain.security.IntegrationKeySpec)
|  java.lang.NullPointerException: Cannot invoke method decrypt() on null object
	at com.unifina.service.EthereumIntegrationKeyService.decryptPrivateKey(EthereumIntegrationKeyService.groovy:128)
	at com.unifina.domain.security.IntegrationKey.jsonMap(IntegrationKey.groovy:55)
	at com.unifina.domain.security.IntegrationKey.toMap(IntegrationKey.groovy:45)
	at com.unifina.domain.security.IntegrationKeySpec.toMap().json for Ethereum integration key(IntegrationKeySpec.groovy:24)
`
I remember wrestling with this months ago. Beans are instantiated differently in test vs production. I think @PostConstruct is ignored when testing or something like that. I remember init() wasn't automatically called in test. I don't understand grails well enough to say exactly why.

test/unit/com/unifina/controller/api/IntegrationKeyApiControllerSpec.groovy
isentropy: `ethereumIntegrationKeyService = mockBean(EthereumIntegrationKeyService, Mock(EthereumIntegrationKeyService))` works instead.

hpihkala:
Whoa! This PR got somehow messed up. Seeing a huge diff with 196 changed files now.

grails-app/services/com/unifina/service/EthereumIntegrationKeyService.groovy
hpihkala: Ok so to keep things reliable, let's delete the `init()` function and get rid of `@PostConstruct` completely. It's ok to initialize `encryptor` lazily, like you're doing in `getStringEncryptor()`. All access to `encryptor` should be via `getStringEncryptor()` anyway.

isentropy:
Yeah, I made a new PR https://github.com/streamr-dev/engine-and-editor/pull/761.

hpihkala:
Replaced by #761

PR #681, opened 2019-10-11
CORE-1631: Fix free product's anon permissions

kare:


grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -141,8 +146,18 @@ class ProductService {
 		}
 
 		Product product = findById(id, currentUser, Permission.Operation.WRITE)
+		if (product.isFree()) {
```
hpihkala: Needs a comment here. Something like "Revoke public read permissions and grant them back after update" (+ why we need to do it?)

PR #683, opened 2019-10-16
Add `dataType` to HTTP module

mondoreale:
Hi guys,

I wanna use the HTTP module for public site's blog section. The problem is that it doesn't handle non-JSON responses very well. I've managed to squeeze the first 13 bytes out of the body of what's clearly an XML but that's all. And it's also kinda weird that our JSON parser lets it be, at all. 13 bytes? A bit confusing. Here's an example of the "13":

Oct-16-2019 10-44-00

_URL: https://medium.com/feed/streamrblog_

—

Anyway, in this PR I add another parameter to the HTTP module aiming at making it more useful. I named it `dataType`. It starts off being set to `auto` which more or less mimics the previous behaviour. Next to it stands: `json`, `text` and `binary`. Here's what I initially expected to see (which now works!):

Oct-16-2019 10-53-22

So, in short, use
- `auto` for the previous behaviour,
- `binary` to get what `toByteArray` gives (binary streams),
- `text` for a utf-8 representation of the response data, and
- `json` for the old default.

Let me know your thoughts.

### DIFF
--- a/src/java/com/unifina/signalpath/remote/AbstractHttpModule.java
+++ b/src/java/com/unifina/signalpath/remote/AbstractHttpModule.java
@@ -345,20 +345,6 @@ public abstract class AbstractHttpModule extends ModuleWithSideEffects implement
 	@Override
 	public void clearState() {}
 
-	public static class DataTypeParameter extends StringParameter {
-		private List<PossibleValue> possibleValues;
-
-		public DataTypeParameter(AbstractSignalPathModule owner, String name, String defaultValue, List<PossibleValue> options) {
-			super(owner, name, defaultValue);
-			possibleValues = options;
-		}
-
-		@Override
-		protected List<PossibleValue> getPossibleValues() {
-			return possibleValues;
-		}
-	}
-
 	public static class VerbParameter extends StringParameter {
 		public VerbParameter(AbstractSignalPathModule owner, String name) {
 			super(owner, name, "POST"); //this.getValueList()[0]);
--- a/src/java/com/unifina/signalpath/remote/AbstractHttpModule.java
+++ b/src/java/com/unifina/signalpath/remote/AbstractHttpModule.java
@@ -345,6 +345,20 @@ public abstract class AbstractHttpModule extends ModuleWithSideEffects implement
 	@Override
 	public void clearState() {}
 
+	public static class DataTypeParameter extends StringParameter {
+		private List<PossibleValue> possibleValues;
+
+		public DataTypeParameter(AbstractSignalPathModule owner, String name, String defaultValue, List<PossibleValue> options) {
+			super(owner, name, defaultValue);
+			possibleValues = options;
+		}
+
+		@Override
+		protected List<PossibleValue> getPossibleValues() {
+			return possibleValues;
+		}
+	}
+
 	public static class VerbParameter extends StringParameter {
 		public VerbParameter(AbstractSignalPathModule owner, String name) {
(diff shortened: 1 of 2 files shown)


src/java/com/unifina/signalpath/remote/Http.java
```diff
@@ -35,6 +35,12 @@
 	private StringParameter URL = new StringParameter(this, "URL", "");
 	private MapParameter headers = new MapParameter(this, "headers");
 	private MapParameter queryParams = new MapParameter(this, "params");
+	private DataTypeParameter dataType = new DataTypeParameter(this, "dataType", "auto", Arrays.asList(
+		new PossibleValue("auto", null),
+		new PossibleValue("binary", "binary"),
+		new PossibleValue("text", "text"),
+		new PossibleValue("json", "json")
```
kare: Should we store content type here also?
Like: `new DataTypePossibleValue("label", "content/type", null)`
Related to my comment down below

src/java/com/unifina/signalpath/remote/Http.java
```diff
@@ -139,15 +146,27 @@ protected void sendOutput(HttpTransaction call) {
...
+
 				if (entity != null) {
-					if (isOctetStream(entity)) {
+					String dt = dataType.getValue();
+
+					if (dt == null) {
+						dt = getDataType(entity);
+					}
+
+					if (dt.equals("binary")) {
```
kare: Here's a nifty Java trick for you: When comparing something with Java's equals(), reverse equals comparison in order to eliminate NullPointerException. Like:
`if ("binary".equals(dt)) { // This works even if dt is null`
Code `dt.equals("")` will throw NullPointerException if dt is null!

It's not necessary here because getDataType(String) will return `!`null.

src/java/com/unifina/signalpath/remote/AbstractHttpModule.java
```diff
@@ -345,6 +345,20 @@ private Propagator getPropagator() {
 	@Override
 	public void clearState() {}
 
+	public static class DataTypeParameter extends StringParameter {
```
kare: This class is only used in sub `Http` class. Maybe it should be moved there?

src/java/com/unifina/signalpath/remote/Http.java
```diff
@@ -170,9 +189,19 @@ protected void sendOutput(HttpTransaction call) {
 		}
 	}
 
-	private static boolean isOctetStream(HttpEntity entity) {
+	private static String getDataType(HttpEntity entity) {
```
kare: Should we only work with standard content types (text/plain, application/xml, etc) here instead of labels (json, binary, etc)?

hpihkala:
A couple of questions/comments actually:

- Won't the `Content-Type` header always dictate how the response should be interpreted? Is there a use case for the user choosing a target format that potentially conflicts with the `Content-Type`?

- Binary doesn't seem useful as an output format, as there are no modules that could do anything with it, and byte arrays are not among the officially supported types.

mondoreale:
Oh btw, medium, the naughty lier:

Screenshot 2019-10-16 23 58 09

mondoreale:
@kare, we good here?

Also, guys, what's your opinion about passing exclusively unparsed data encoded as utf-8 to the `data` output? – something I talk about in [my above comment](https://github.com/streamr-dev/engine-and-editor/pull/683#pullrequestreview-302893860).

mondoreale:
Alright then. I'mma merge it in shortly, as is. Thanks guys!

PR #682, opened 2019-10-14
CPS-46: ACCEPTED JoinRequests don't get sent to joinPartStream

kare:

PR #685, opened 2019-10-17
Logging tweaks

hpihkala:
- Reduce noise from `AbstractHttpModule` and `ContractEventPoller`
- Cleanup code related to exception logging
- Only log unexpected exceptions (status code `500` basically) to reduce noise
- Add debug logging to `CommunityJoinRequestService` and `StreamrClientService`
- Turn on debug logging for above classes and `StreamrClient` to debug join issues in production

This PR includes PR #682.

### DIFF
--- a/grails-app/controllers/com/unifina/controller/api/ErrorController.groovy
+++ b/grails-app/controllers/com/unifina/controller/api/ErrorController.groovy
@@ -63,7 +63,7 @@ class ErrorController {
 		}
 
 		// Log internal errors
-		if (apiError.statusCode == 500) {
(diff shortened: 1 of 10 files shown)


grails-app/controllers/com/unifina/controller/api/ErrorController.groovy
```diff
@@ -66,6 +62,11 @@ class ErrorController {
 			apiError = new ApiError(500, exception.class.simpleName, exception.getMessage())
 		}
 
+		// Log internal errors
+		if (apiError.statusCode == 500) {
```
kare: Catch all 5xx server error codes here?

PR #686, opened 2019-10-17
Fix UI channel transitive permissions

hpihkala:
If a user has certain permissions to a canvas, he should also have those same permissions to the UI channel streams of that canvas.

Currently, the `/streams/:streamId/permissions/me` endpoint does not list those transitive permissions for UI channel streams. See the [JIRA ticket](https://streamr.atlassian.net/browse/CORE-1825).

I started working on a fix, but got stuck with some weird behavior of the unit test. Unrelated permissions were being returned by the criteria query. @harbu or @kare please take this over and see if you can get the test working. I probably can't put more work into this before MozFest.

### DIFF
--- a/grails-app/domain/com/unifina/domain/security/Permission.groovy
+++ b/grails-app/domain/com/unifina/domain/security/Permission.groovy
@@ -149,4 +149,9 @@ class Permission {
 		}
 		return map
 	}
+
+	@Override
+	String toString() {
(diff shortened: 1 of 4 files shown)


hpihkala:
@kare could you help debug the new test?

hpihkala:
@kare thanks!

PR #690, opened 2019-10-21
Get secret from api

jtakalai:
API endpoint /communities/.../secrets/id returns the actual secret too, now

useful for autogenerated secrets. How could you otherwise use them at all?


grails-app/domain/com/unifina/domain/community/CommunitySecret.groovy
```diff
@@ -23,15 +23,17 @@ class CommunitySecret {
...
-		//  We probably need to be able to return the secret in the object?
-		//  Especially if it's generated server-side.
-		//  The API should of course only serve the secrets to the community admin.
-		return [
-			id: id,
-			name: name,
-			communityAddress: communityAddress,
-		]
-	}
+	Map toMap() { [
```
hpihkala: `toMap(boolean includeSecret = false)` would be more DRY and still be backwards compatible?

.gitignore
```diff
@@ -14,3 +14,4 @@ replay_pid*
 /.slcache/
 /tomcat.8081/
 /rest-e2e-tests/node_modules/
+web-app/webpack-bundles
```
hpihkala: No such directory exists anymore, and no need to add it to `.gitignore`. If it exists on your machine for some reason, just delete it.

grails-app/services/com/unifina/service/CommunitySecretService.groovy
```diff
@@ -33,7 +33,12 @@ class CommunitySecretService {
 		if (result == null) {
 			throw new NotFoundException("Community secret not found")
 		}
-		result.name = cmd.name
+		if (cmd.name) {
+			result.name = cmd.name
+		}
+		if (cmd.secret) {
+			result.secret = cmd.secret
```
hpihkala: Question: Why do we allow the user to set the secret, instead of always generating it? (also applies to line 19, but can't comment there)

jtakalai: it's a matter of order; suppose you'd have to create the secret and deploy it (maybe to an IoT gadget), and only later enable the secret in EE. Not sure if this is a real use case.

jtakalai:
TODO: PUT should not be able to change the secret

grails-app/services/com/unifina/service/CommunitySecretService.groovy
jtakalai: I'll throw an error then instead. Better than silently 200 and not change the secret IMO.

grails-app/services/com/unifina/service/CommunitySecretService.groovy
```diff
@@ -33,7 +34,12 @@ class CommunitySecretService {
 		if (result == null) {
 			throw new NotFoundException("Community secret not found")
 		}
-		result.name = cmd.name
+		if (cmd.name) {
+			result.name = cmd.name
+		}
+		if (cmd.secret) {
+			throw new BadRequestException("Can't change the secret, please create a new community secret")
```
kare: Needs Swagger update

test/integration/com/unifina/service/CommunitySecretServiceIntegrationSpec.groovy
```diff
@@ -116,6 +118,59 @@ class CommunitySecretServiceIntegrationSpec extends Specification {
...
+	void "update() only secret"() {
+		setup:
+		CommunitySecret s1 = new CommunitySecret(
+			name: "secret 1",
+			secret: "secret#1",
+			communityAddress: communityAddress,
+		)
+		s1.save(validate: true, failOnError: true)
+		CommunitySecretCommand cmd = new CommunitySecretCommand(
+			secret: "new secret",
```
kare: Could we drop the secret field from CommunitySecretCommand?

jtakalai: that would make it like apikey, sure. Then you couldn't input your own secret, I guess? I think I'll keep it like it is for now

grails-app/services/com/unifina/service/CommunitySecretService.groovy
jtakalai: Swagger is updated now. That parameter is not in the spec, so it returns the 400 (which is already in the spec).

PR #694, opened 2019-10-31
CORE-1842: Fix product validation on update/PUT

kare:


src/groovy/com/unifina/api/UpdateProductCommand.groovy
```diff
@@ -47,14 +47,7 @@ class UpdateProductCommand {
...
-		name(blank: false)
-		description(blank: false)
-		previewStream(nullable: true)
-		previewConfigJson(nullable: true)
-		pendingChanges(nullable: true)
-
-		ownerAddress(nullable: true, validator: Product.isEthereumAddressOrIsNull)
-		beneficiaryAddress(nullable: true, validator: Product.isEthereumAddressOrIsNull)
+		importFrom(Product)
 		pricePerSecond(nullable: true)
```
hpihkala: Hmm.. How about these? Or should we list all the `onChainFields` here as `nullable`?  Why do we now keep some of them?

PR #696, opened 2019-10-31
some bugfixes for getEvents

isentropy:
1. fixed bug where 0 parameter event was always sent, even if another event was triggered.
2. RuntimeException caught in WebsocketEthereumJsonRpc.openConnectionRetryIfFail(). DNS lookup failure causes a RuntimeException, which wasn't caught
3. InterruptedException caught when sleeping between connection retries. I'm not sure why thread was interrupted, but I've seen this in E&E.
4. No RuntimeExceptions thrown when writing to closed WS. WebsocketEthereumJsonRpc should reopen itself, but event poller shouldn't throw RuntimeException

I'll check out switching to Web3j WS connector for events polling. Made a ticket here: https://streamr.atlassian.net/browse/CORE-1813. Probably healthier to replace this in the long term.


hpihkala:
Could @jtakalai please review this? Removing myself.

Maybe my five cents: what about tests? If there was an issue that was fixed, can we add a test that would have failed before these changes?

isentropy:
I can add a GetEventsSpec unit test class. There isn't one atm. The first issue is straightforward to test for in unit test, but test the other reconnection issues may require integration test with a real, stateful ws geth rpc. I'll discuss with Juuso.

isentropy:
BTW issues 1-3 are somewhat exotic:

I found #1 by reviewing code. Didn't actually discover by testing.

#2 only occurs when you disconnect internet long enough for DNS cache to be wiped.

I saw #3 only once. Not sure why thread gets interrupted.

@jtakalai lets discuss on call. I would really like to replace this homegrown RPC code with Web3j in any case.

src/java/com/unifina/signalpath/blockchain/GetEvents.java
```diff
@@ -212,10 +212,10 @@ protected void displayEventsFromLogs(LogsResult lr) {
 						}
 						convertAndSend(output, value);
 					}
+				} else{
+					// events with no arguments just get a true sent as a sign of event being triggered
+					eventOutputs.get(0).send(Boolean.TRUE);
```
jtakalai: Can you elaborate what this fixes? Looking at the code, there are two cases: `abiEvent.inputs` is empty, or not. If empty, just do the boolean output; if not, do proper value outputs. Switching for-loop over if-branch doesn't seem to change much in the way of handling those, thought it DOES hilight a rather crucial problem with the current setup: what if there are several events of given type in the block? Right now all would end up in `lr.logs` and later ones would overwrite earlier ones. Instead we should store those in a list and propagate them one by one, in other words: do the parsing already in onEvent, and then `enqueueEvent` the logs individually, preferably already in ready-to-output form.

Also I noticed `valueList==null` on line 194 is always true because `extractEventParameters` doesn't seem to ever return null. Should it check for zero length instead?

isentropy: Before the fix, any event with 0 params in always triggered by any log, even one for the wrong event. By putting the 0 param event within this loop
`//valueList contains only the events of type abiEvent
for (EventValues ev : valueList) {`
``
we insure that it is only triggered when we get an EventValues that matches that 0 param event. You can see the bug and fix by locally testing with this contract:
`
pragma solidity ^0.4.24;
contract StockTicker{
	address public owner;
	event PriceUpdated(
		string symbol,
		uint64 indexed timestamp,
		uint price_usdwei,
		string indexed indexed_symbol
	);
	event BlankEvent();
	constructor() public{
		owner = msg.sender;
	}
	function blank() public{
		emit BlankEvent();
	}
	function updatePrice(string memory symbol, uint64 timestamp, uint price_usdwei) public{
		require(msg.sender == owner, "only owner can post stock data");
		emit PriceUpdated(symbol, timestamp, price_usdwei, symbol);
	}
}
`

before the fix, calling updatePrice() shows BlankEvent triggered. After fix, it does not.

isentropy: Yes re line 194. That null check can be removed.

hpihkala:
@isentropy what's the status of this PR please?

hpihkala:
Cleaning up this old PR, most of it was merged in #703

PR #699, opened 2019-11-02
Throw away deprecated Ethereum modules that depended on streamr-web3

jtakalai:
If someone's canvas uses them still, it won't work anyway


jtakalai:
TODO: migrations to replace those modules in old canvases with new modules

PR #701, opened 2019-11-05
Core 1839 inbox stream permissions fix

mthambipillai:


hpihkala:
Does this solve the bloat issue? How? Can a test be made to verify?

mthambipillai:
When a subscription is updated in `SubscriptionService.updateSubscriptionAndLinkedPermissions`, the old permissions are first deleted and then the new permissions are added. When deleting, using `.delete()`, inbox stream permissions were not deleted, but they were added in the next step through `systemGrant()`. So every subscription update was creating duplicate inbox stream permissions.
Unit tests have been updated to check that `systemRevoke()` is called (and not just `delete()`).

hpihkala:
@mthambipillai I think much of the bloat problem we were seeing was due to duplicate "parent" permissions, i.e. the same users having read and write permissions multiple times in the database. This lead to large multipliers in front of the expected `R*W*2` inbox stream permissions created for `R` readers and `W` writers.

Nowadays re-joining a community product won't generate duplicate permissions, this was fixed in [this commit](https://github.com/streamr-dev/engine-and-editor/commit/28d4a7686fc4afd9e62a32a7a4063cbf745e6b99). That should help limit the number of duplicate write permissions.

One source of duplicate read permissions still exists: expired subscriptions. I created a test to show that even for expired read permissions, the child inbox stream permissions are being created (e.g. when new writers join). This will quickly lead to permission bloat. @mthambipillai could you take a look if this could be fixed?

mthambipillai:
JIRA ticket created: https://streamr.atlassian.net/browse/CORE-1861

PR #703, opened 2019-11-05
Getevent bugfixes minimal pr 20191105

isentropy:
This PR has the bugfixes from https://github.com/streamr-dev/engine-and-editor/pull/696, without the Ethereum modules refactor that Juuso did. 

Contains the following bugfixes:
1. fixed bug where 0 parameter event was always sent, even if another event was triggered.
2. RuntimeException caught in WebsocketEthereumJsonRpc.openConnectionRetryIfFail(). DNS lookup failure causes a RuntimeException, which wasn't caught.
3. InterruptedException caught when sleeping between connection retries. I'm not sure why thread was interrupted, but I've seen this in E&E.
4. No RuntimeExceptions thrown when polling to closed WS. This happens when ContractEventPoller polls the RPC while the websocket is closed and attempting reopen.

### DIFF
--- a/src/java/com/unifina/signalpath/blockchain/WebsocketEthereumJsonRpc.java
+++ b/src/java/com/unifina/signalpath/blockchain/WebsocketEthereumJsonRpc.java
@@ -112,7 +112,7 @@ public class WebsocketEthereumJsonRpc extends EthereumJsonRpc {
 					log.error("Thread was interrupted while sleeping.");
 					//why is this thread interrupted in E&E when sleeping?
 					log.error(ie.getMessage());
-					Thread.interrupted();
+					Thread.currentThread().interrupt();
 				}
 				continue;
 			}
--- a/src/java/com/unifina/signalpath/blockchain/WebsocketEthereumJsonRpc.java
+++ b/src/java/com/unifina/signalpath/blockchain/WebsocketEthereumJsonRpc.java
@@ -114,6 +114,7 @@ public class WebsocketEthereumJsonRpc extends EthereumJsonRpc {
 					log.error("Thread was interrupted while sleeping.");
 					//why is this thread interrupted in E&E when sleeping?
 					log.error(ie.getMessage());
+					Thread.interrupted();
 				}
 				continue;
 			}
--- a/src/java/com/unifina/signalpath/blockchain/WebsocketEthereumJsonRpc.java
+++ b/src/java/com/unifina/signalpath/blockchain/WebsocketEthereumJsonRpc.java
@@ -99,8 +99,6 @@ public class WebsocketEthereumJsonRpc extends EthereumJsonRpc {
 			log.info("Trying to establish websocket connection to " + url + ". Attempt number " + attempts);
 			try {
 				openConnection();
-				//in the case of ContractEventPoller, handler.init() installs filter
-				handler.init();
 			} catch (URISyntaxException e) {
 				log.error(e);
 				return false;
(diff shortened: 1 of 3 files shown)


src/java/com/unifina/signalpath/blockchain/WebsocketEthereumJsonRpc.java
```diff
@@ -112,10 +104,17 @@ public boolean openConnectionRetryIfFail() throws InterruptedException {
...
+			} catch (IOException | DeploymentException | RuntimeException e) {
 				log.error(e);
 				log.info("Waiting " + attemptReconnectionWaittimeMs + "ms");
-				Thread.sleep(attemptReconnectionWaittimeMs);
+				try {
+					Thread.sleep(attemptReconnectionWaittimeMs);
+				}
+				catch(InterruptedException ie){
+					log.error("Thread was interrupted while sleeping.");
+					//why is this thread interrupted in E&E when sleeping?
```
kare: Try `Thread.currentThread().interrupt();` as explained here in detail: https://stackoverflow.com/questions/1087475/when-does-javas-thread-sleep-throw-interruptedexception

isentropy: OK, thanks. Maybe I'll keep throwing the InterruptedException then.

isentropy:
API test is timing out. Trying to figure out why. I tried restarting test and it happens repeatedly. Hold off on merge until CI works of course.

src/java/com/unifina/signalpath/blockchain/WebsocketEthereumJsonRpc.java
jtakalai: Hey this seems a very good point: maybe add the interrupt() after log. Not interrupting the thread might leave the app hanging when it should shut down or something? Should at least try if it works with the `interrupt()`. If it's not broken with it, then please add it. If `interrupt()` breaks it then... maybe just swallow it.

isentropy: i'll test. I don't think the InterruptedException is a major issue in any case

isentropy:
After restarting CI about 4-5 times, it magically works.

src/java/com/unifina/signalpath/blockchain/WebsocketEthereumJsonRpc.java
jtakalai: Here we should also set `attempts = attemptReconnectionMaxTries` to break out from the loop (or maybe `break`), and the thing will then throw a RuntimeException.

kare: If there are multiple threads then InterruptedException should be handled. I don't think that logging exception adds value here but calling interrupt() does.

isentropy:
I added Thead.interrupted(). I will test by letting editor run for a while with internet disconnected. If you're ok with the PR and tests pass, I'll merge tomorrow. Thanks for your help @kare

isentropy:
Well I think breaking is effectively the same as throwing InterruptedException: the module will stop running. I only saw the InterruptedException while sleeping once, and I'm not even sure why it happened. It might have been a side effect of RuntimeException throws by the polling issue. Let me see how editor handles some long disconnects. I think it's an exotic issue. I can just throw it as before if you prefer.

isentropy:
It's working well. I left internet off for 2hrs, restarted connection, and GetEvents recovered without any issues. I'll merge this tomorrow morning if there are no objections.

src/java/com/unifina/signalpath/blockchain/WebsocketEthereumJsonRpc.java
jtakalai: according to the text, the `interrupt()` is kind of a simple fix that shouldn't break anything (if all it does is set a flag), so it should be added; of course if it DOES break things, then add a TODO and think about it later.

PR #704, opened 2019-11-06
CORE-1840: Fix timer drift bug

kare:


test/unit/com/unifina/datasource/DataSourceEventQueueSpec.groovy
```diff
@@ -147,4 +146,43 @@ class DataSourceEventQueueSpec extends Specification {
 		})
 	}
 
+	void "timer drift"() {
```
hpihkala: Years from now, it might be hard to remember what is being tested. Maybe something like:

```suggestion
	void "every full second between subsequent event timestamps is being reported"() {
```

test/unit/com/unifina/datasource/DataSourceEventQueueSpec.groovy
```diff
@@ -147,4 +146,43 @@ class DataSourceEventQueueSpec extends Specification {
...
+				}
+			}))
+			queue.enqueue(new Event<Integer>(2, new Date(20000), new Consumer<Integer>() {
+				@Override
+				void accept(Integer integer) {
+					eventsProcessed++
+				}
+			}))
+		}
+
```
hpihkala: Would be good to test that each second in between these two events is being reported. So you could additionally do something like (untested code):

```
ITimeListener timeListener = Mock(ITimeListener)
timeListener.tickRateInSec() >> 1
queue.addTimeListener(timeListener)
```

And then assert that the `timeListener.setTime(...)` gets called N times (I guess 11 times in this case)

The timestamps it should be called with are:

```
10000
11000
...
20000
```

test/unit/com/unifina/datasource/DataSourceEventQueueSpec.groovy
```diff
@@ -147,4 +146,43 @@ class DataSourceEventQueueSpec extends Specification {
...
+		then:
+		new PollingConditions().within(10, {
+			eventsProcessed == 2
```
hpihkala: Possible mistake here. Code is equivalent to

```
		new PollingConditions().within(10, {
			eventsProcessed == 2    // THE VALUE OF THIS STATEMENT IS IGNORED!
			return !producerThread.isAlive() // ONLY THIS IS TESTED
		})
```

Should change to

```
		new PollingConditions().within(10, {
			eventsProcessed == 2 && !producerThread.isAlive()
		})
```

test/unit/com/unifina/datasource/DataSourceEventQueueSpec.groovy
```diff
@@ -147,4 +146,43 @@ class DataSourceEventQueueSpec extends Specification {
...
+		}
+
+		Thread producerThread = Thread.start {
+			queue.enqueue(new Event<Integer>(1, new Date(10000), new Consumer<Integer>() {
+				@Override
+				void accept(Integer integer) {
+					eventsProcessed++
+				}
+			}))
+			queue.enqueue(new Event<Integer>(2, new Date(20000), new Consumer<Integer>() {
```
hpihkala: Add slightly more testing power to not have an even second as the end event timestamp

```suggestion
			queue.enqueue(new Event<Integer>(2, new Date(20100), new Consumer<Integer>() {
```

test/unit/com/unifina/datasource/DataSourceEventQueueSpec.groovy
```diff
@@ -147,4 +146,43 @@ class DataSourceEventQueueSpec extends Specification {
...
+				}
+			}))
+		}
+
+		then:
+		new PollingConditions().within(10, {
+			eventsProcessed == 2
+			!producerThread.isAlive()
+		})
+		queue.getLastReportedClockTick() == 20000
```
hpihkala: Internal API, would be better to check via attaching the `ITimeListener` , in which case this line is not needed

PR #705, opened 2019-11-06
Update REST API test dependencies

kare:
Fixes #698

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ jdk:
 services:
 - docker
 node_js:
-- "8.12.0"
+- "10.16.3"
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -5,7 +5,7 @@ jdk:
 services:
 - docker
 node_js:
-- "8.9.3"
```
harbu: Would it make sense to upgrade Node.Js to 10 or higher? LTS for 8 is ending in January: https://nodejs.org/en/about/releases/

kare: @harbu 
```
% node --version
v10.16.3
% npm --version
6.9.0
```

PR #707, opened 2019-11-12
Add uiChannel & canvas restart tests

timoxley:
1. Adds a test checking that uiChannel data is flowing by adding a Table to the existing Stream->Multiply->Sum test canvas.
2. Tests that data still flows after rapidly stopping and starting a canvas.

Variants of these tests were failing over the past few days, seem to be passing consistently now with latest docker images.

3. ~Also adds a failing test for resend data. It appears data flows but uiChannel resends don't seem to work. UI resends seem to be working just fine in the actual app, albeit with a clock->table, so it's possible my test is not correct, please verify my test makes sense @hpihkala.~

Ah nvm about the resend UI issue, my test didn't make sense, its not working because no resend parameters are set on the subscription, it doesn't magically set the resend parameters when you change the options on the module itself. Of course. ~~Will fix that,~~ fixed but I believe the clock canvas test failure is still valid.

~I also want to create a test that uses a Clock but this is now a bit problematic if the test is ever not shut down gracefully since the clock canvas(es) will keep running forever in the background  filling the already noisy backend logs with distracting messages.~ Added this test anyway.

4. Added a test for a Clock -> Table that ensures at least one new row message arrives on the uiChannel after starting the canvas. This test fails intermittently, so I have set it up to repeat multiple times e.g.

image

### DIFF
--- a/rest-e2e-tests/canvas-api.test.js
+++ b/rest-e2e-tests/canvas-api.test.js
@@ -209,6 +209,7 @@ describe('Canvas API', function() {
 
     describe('restarting canvas', () => {
         const messages = []
+        let resentMessages
         let subscription
         before('cycle start/stop, subscribe, start', async () => {
             let done
@@ -231,6 +232,10 @@ describe('Canvas API', function() {
                 messages.push(msg)
             })
 
+            subscription.once('resent', () => {
+                resentMessages = messages.slice()
+            })
+
             // wait for subscription before starting again
             subscription.once('subscribed', async () => {
                 // restart for second time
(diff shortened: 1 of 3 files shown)


hpihkala:
@harbu @mirpo Tim managed to write a test for what's often happening in the frontend. The test still involves canvases so it's not as isolated as it could be, but this seems like a network issue, because upon browser refresh (subscriber reconnects) everything works. 

This is the thing I mentioned last week, which could be due to some race condition where a publisher and subscriber for the same stream are appearing at the same time. Could you see if you can make the test fail locally (the test logic basically starts on line `315`), and investigate what's happening in the network?

hpihkala:
Included in #714

PR #708, opened 2019-11-12
CORE-1832: Add transitive UI channel read-permissions via dashboards

kare:


grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -139,6 +138,14 @@ class PermissionService {
 			}
 		}.toList()
 
+		if (resource instanceof Stream && resource.uiChannel && resource.uiChannelCanvas != null && resource.uiChannelPath != null) {
```
hpihkala: Is this basically the same condition as line `150`? They could be combined. 

(This one has some additional conditions which we can keep, but basically `uiChannelCanvas` and `uiChannelPath` should always be set if `resource.uiChannel` is true.)

grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -147,6 +154,26 @@ class PermissionService {
 		return directPermissions
 	}
 
+	private Permission hasTransitiveDashboardPermissions(Canvas canvas, Userish userish) {
+		List<DashboardItem> items = DashboardItem.findAllByCanvas(canvas)
+		Permission permission = Permission.withCriteria(uniqueResult: true) {
```
hpihkala: Will this throw if the result is not unique? I faintly remember there might be some trap like that

grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -147,6 +154,26 @@ class PermissionService {
 		return directPermissions
 	}
 
+	private Permission hasTransitiveDashboardPermissions(Canvas canvas, Userish userish) {
+		List<DashboardItem> items = DashboardItem.findAllByCanvas(canvas)
```
hpihkala: Result set could theoretically be large, but in practice it's unlikely to be. Would be better to do it all in one query using a join, but that said, I think the current approach will work for now & figuring out how to do it correctly via Hibernate criteria queries might take too much time vs. the benefits.

grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -533,11 +560,19 @@ class PermissionService {
...
 
 		// Special case of UI channels: they inherit permissions from the associated canvas
-		if (p.empty && resource instanceof Stream && resource.uiChannel) {
-			return hasPermission(userish, resource.uiChannelCanvas, op)
+		if (p.isEmpty() && resource instanceof Stream && resource.uiChannel) {
+			if (hasPermission(userish, resource.uiChannelCanvas, op)) {
+				return true
+			}
+		}
+		if (p.isEmpty() && resource instanceof Stream && resource.uiChannel) {
```
hpihkala: Same condition as line 563, can be combined

test/integration/com/unifina/service/PermissionServiceIntegrationSpec.groovy
```diff
@@ -310,4 +342,12 @@ class PermissionServiceIntegrationSpec extends IntegrationSpec {
 		service.canWrite(me, stream)
 		service.canShare(me, stream)
 	}
+
+	void "getPermissionsTo(resource, userish) returns correct UI channel read permissions via associated dashboard"() {
+		service.systemGrantAll(me, dashboard)
+
+		expect:
+		service.getPermissionsTo(uiChannelStream, me).size() == 1
```
hpihkala: Minor sanity check improvement: could check that there is a single item AND that single item is a `read` permission

PR #711, opened 2019-11-19
NET-88: fix "clock-table" bug

harbu:
Fix "clock-table" bug.

The bug is caused by invoking `pushToUiChannel` too early in a module, i.e., before `DataSourceEventQueue#initTimeReporting` is executed. Because `globals.time` is used as timestamp in published messages and `globals.time` is adjusted backwards in `initTimeReporting`, if a module uses `pushToUiChannel` in both its `sendOutput()` and `initialize()` implementations, the resulting chain of `(timestamp, sequenceNo)`pairs isn't necessarily monotonically increasing, which is a requirement of message chains in the Streamr network.

Fixed instances where a module would try to `pushToUiChannel` in `initialize`.

Fixes issue that can be replicated by #707. I tried running the test case defined there at least five times without problems against this branch.

### DIFF
--- a/test/unit/com/unifina/signalpath/charts/TimeSeriesChartSpec.groovy
+++ b/test/unit/com/unifina/signalpath/charts/TimeSeriesChartSpec.groovy
@@ -62,6 +62,9 @@ class TimeSeriesChartSpec extends UiChannelMockingSpecification {
 		new ModuleTestHelper.Builder(module, inputValues, outputValues)
 			.timeToFurtherPerIteration(1000)
 			.uiChannelMessages(channelMessages, getSentMessagesByStreamId())
+			.afterEachTestCase {
+				module.headersSent = false
+			}
 			.test()
 	}
 }
--- a/src/java/com/unifina/signalpath/charts/Chart.java
+++ b/src/java/com/unifina/signalpath/charts/Chart.java
@@ -9,6 +9,7 @@ import java.util.Map;
 
 public abstract class Chart extends ModuleWithUI {
 	private String dataGrouping = "min/max";
+	protected boolean headersSent = false;
 
 	@Override
 	public void init() {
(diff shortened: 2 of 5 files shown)


test/unit/com/unifina/signalpath/charts/TimeSeriesChartSpec.groovy
```diff
@@ -62,6 +62,9 @@ class TimeSeriesChartSpec extends UiChannelMockingSpecification {
 		new ModuleTestHelper.Builder(module, inputValues, outputValues)
 			.timeToFurtherPerIteration(1000)
 			.uiChannelMessages(channelMessages, getSentMessagesByStreamId())
+			.afterEachTestCase {
+				module.headersSent = false
```
harbu: `Chart` has set `canClearState = false`. Thus this manual invocation after each test case is necessary for the header message to be sent.

harbu:
@timoxley agree strongly. I was going to implement the 1st bullet of your list but opt'd to go for a minimal solution due to time constraints.

Some other solution ideas (not necessarily advocating just food for thought):
- You could also leverage the type system to not have `pushToUiChannel` even available during configuration. This would split modules into two classes: their configuration and runtime counter-parts. Of course this would require a massive refactoring and essentially require every module to have two classes, which would indeed be painful. (Granted utilizing anonymous inner class and Java 8 functional programming facilities could make this less painful.)
- Is pushing headers into the same stream that contains the actual data points necessarily the "right" place? Often in Streamr streams we expect the data points to adhere to a fixed schema. We could use, e.g., the runtime request system to fetch the headers of a Chart / Table whenever necessary.

hpihkala:
Should we dig more into the root cause? Why is `globals.time` being set to a smaller value than what it gets initialized with, and is it that necessary? If that gets fixed, then the modules can't do anything "wrong" in the first place.

- What value is `globals.time` set to initially? This is probably the value it has when `initialize` gets called on modules.
- What timestamp does the first handled event have? Is that some "startEvent", the timestamp of which we set, or is it just whatever event first comes from a stream or a clock?

Something like this may be happening currently (just guessing off the top of my head):
- `globals.time` gets set to current system time initially (in case of realtime canvases)
- A start event with that timestamp gets inserted as the first event in the event queue
- When that start event gets handled, the time reporting is initialized from the previous even second

This sequence would lead to `globals.time` always taking a step backwards in the beginning, unless the initial value of `globals.time` happens to be an even second. 

It would be fairly easy to fix by ensuring that the first value of `globals.time` is less than or equal to the time that the time reporting gets initialized with. Could for example round down the initial value of `globals.time` to the even second.

hpihkala:
Tried to find and fix the root cause, see #714

harbu:
I'll close this since we've agreed to go with solution in PR #714

PR #714, opened 2019-11-25
Fix time reporting in DataSourceEventQueue

hpihkala:
Timestamps smaller than the initial value of `globals.time` were being reported to `ITimeListener`s in realtime mode. This lead to problems in case time listeners (such as the `Clock` module) read time when the module is initialized, because the first reported clock tick was less than the time at initialization.

This PR includes a slight refactoring to start the time ticking from the next full second that follows the initial value of `globals.time`.

This PR is an alternative solution to #711 by @harbu, which requires modules to guard against the bug in time reporting. Assuming that this PR does indeed fix the underlying bug, the modules don't need to guard against this issue anymore.

@timoxley 's tests added in #707 seem to pass with this branch. Would be good if you guys could do some further validation.

### DIFF
--- a/rest-e2e-tests/canvas-api.test.js
+++ b/rest-e2e-tests/canvas-api.test.js
@@ -12,6 +12,10 @@ const Streamr = initStreamrApi(REST_URL, LOGGING_ENABLED)
 
 const TIMEOUT = 30 * 1000
 
+const NUM_MESSAGES = 50
+
+const WAIT_TIME = 15000
+
 const pollCondition = async (condition, timeout = TIMEOUT, interval = 100) => {
     let timeElapsed = 0
     let result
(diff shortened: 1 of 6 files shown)


hpihkala:
I'm merging in the additional tests from @timoxley's branch into this PR

hpihkala:
Almost there. @timoxley could you help debug the new rest-e2e-tests? They pass on my machine but one of them failed on Travis, as seen above. Could be just flakiness, but I didn't re-run in order to show it to you. Add more timeout? :)

timoxley:
@hpihkala Luckily, the same test was failing on my machine. More timeout helped, but so that the test is less dependent on time guesswork I've tweaked it so now it waits for the correct message to arrive, up to the suite's test timeout, rather than sleeping for a  number of seconds before checking whether a message arrived. This gets things passing consistently on my machine, pushed the change into this PR and now 🤞 it also works in CI.

timoxley:
Ahh, nope, test doesn't work in CI and turns out it doesn't work reliably locally either

timoxley:
@hpihkala I'm seeing the same failing behaviour with the repeated restart tests on the `master` branch, so at least this issue isn’t introduced by these changes, it already exists.

Note: the last few tests in the screenshot below "should" fail on `master` because it doesn't include the fix in this PR.
image

timoxley:
Summary of the failing test:

1. starts a canvas
2. stops it
3. subscribes to a uiChannel
4. waits for subscription
5. starts the canvas again
6. publishes a message
7. waits for the message to come though

Timeout seems to occur when it's waiting for the new message to appear.

hpihkala:
Fixed some of the flakiness by adding a delay to allow the subscriptions of the canvas to start working before we publish new messages to it.

Still getting this weird issue sometimes. Everything seems to be correct, but the client decides (mid-resend!) that there shouldn't be a resend in progress.

Screen Shot 2019-11-29 at 18 27 53

hpihkala:
I updated my `broker` docker image, and that might have fixed the above remaining issue. Can't get this to fail locally anymore. Merging. 

While playing around I discovered a network issue which makes these (and other) tests fail after restarting broker nodes individually in the local env. Filing a `NET` ticket about that.

PR #715, opened 2019-11-26
CORE-1850: Setup HTTP connection pool for CPS proxy

kare:
I've manually tested this: https://streamr.atlassian.net/browse/CORE-1850

Could use a little more testing or different approach.


hpihkala:
@kare Question: Why did you port the service from Groovy to Java?

kare:
@hpihkala I couldn't get constructor injection to work with Grails. See resources.groovy

hpihkala:
@kare The normal pattern is to read the config in the class (instead of passing them in) from the automatically injected `grailsApplication` bean. It is injected as long as the variable is defined. 

A few examples of reading config entries via injected `grailsApplication` bean in a service: 
- [in MetricsService](https://github.com/streamr-dev/engine-and-editor/blob/master/grails-app/services/com/unifina/service/MetricsService.groovy#L41)
- [in ClusterService](https://github.com/streamr-dev/engine-and-editor/blob/master/grails-app/services/com/unifina/service/ClusterService.groovy#L69)

You can also access config via the static `Holders` object, but this usually only used in non-bean objects such as POJOs.

kare:
@hpihkala Ok, but I use that constructor in the last test called: "test socket read timeout". Should I convert it back to Groovy?

test/unit/com/unifina/service/CommunityOperatorServiceSpec.groovy
```diff
@@ -127,17 +127,4 @@ class CommunityOperatorServiceImplSpec extends Specification {
 			assert results.get(i).response.statusCode == 200
 		}
 	}
-
-	void "test socket read timeout"() {
```
hpihkala: Should we keep this test? It seems useful.

If you want to manipulate the config for the purposes of the unit test (set a lower timeout, etc.), you can certainly do so, see an example [here](https://github.com/streamr-dev/engine-and-editor/blob/master/test/unit/com/unifina/service/NodeServiceSpec.groovy#L18). 

NOTE: It may require you to switch to accessing the config via `grailsApplication` instead of the static `Holders` object. In my experience, it's harder to manipulate the static config in unit tests (it may not change even if you manipulate the config in the unit test). For this reason (better testability), accessing config via `grailsApplication` is preferred.

PR #718, opened 2019-12-02
update streamr-client-java to 1.2.2

harbu:


harbu:
Re-ran tests, passed. Merging.

PR #724, opened 2019-12-13
fix: DEVOPS-181 AWS Image upload for docker:

mikhaelsantos:
* Refelect env changes to CATALINA_OPTS


mikhaelsantos:
> These properties are already difficult enough to manage. Let's try to find an approach that leaves us easy to read diffs.
> 
> Is it possible to set CATALINA_OPTS variable on CMD line?
> Like: CMD ["sh" .... $CATALINA_OPTS catalina.sh run"]

No. CATALINA_OPTS doesn't refresh if o§ther vars are changed. Check issue for the description of the problem in more detail

PR #726, opened 2019-12-18
CORE-1857: Add validation for Stream.partitions

kare:


grails-app/domain/com/unifina/domain/data/Stream.groovy
```diff
@@ -9,11 +9,19 @@ import grails.converters.JSON
 import groovy.transform.CompileStatic
 
 class Stream implements Comparable {
+	public final static Integer DEFAULT_PARTITIONS = 1
 	public final static String DEFAULT_NAME = "Untitled Stream"
+	public final static boolean DEFAULT_UI_CHANNEL = false
+	public final static boolean DEFAULT_INBOX = false
+	public final static boolean DEFAULT_REQUIRE_SIGNED_DATA = false
```
hpihkala: Hmm, isn't this unnecessary boilerplate? Was there a problem with giving the default values on variable introduction?

grails-app/domain/com/unifina/domain/data/Stream.groovy
```diff
@@ -51,6 +59,7 @@ class Stream implements Comparable {
 	static belongsTo = Product
 
 	static constraints = {
+		partitions(nullable: false, min: 1, max: 100)
```
hpihkala: Good range! 👍

grails-app/domain/com/unifina/domain/data/Stream.groovy
kare: @hpihkala No problem, just thought that it could be useful in unit tests, but then again maybe not.

PR #732, opened 2020-01-14
hotfix: reduce CassandraService query from 1 year to 1 week

harbu:
Currently these queries seem to timeout a lot in Cassandra and may cause impact on broker service reliability. Let's test out reducing the result set of the query if this helps with timeouts.


harbu:
Good points. I will not, however, implement these in this PR since I just need this quick change to gauge how E&E is affecting Cassandra performance.

PR #735, opened 2020-02-05
CORE-1833: New resource permissions

kare:
- Tests pass when ran with custom broker


grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -184,18 +184,18 @@ class ProductService {
 
 	void addStreamToProduct(Product product, Stream stream, SecUser currentUser)
 			throws ValidationException, NotPermittedException {
-		permissionService.verifyShare(currentUser, stream)
+		permissionService.verifyShareStream(currentUser, stream)
```
hpihkala: We know that it's a `Stream`, so `permissionService.verify(currentUser, stream, Operation.STREAM_SHARE)` avoids the need for resource-specific methods in `PermissionService` (see earlier comment).

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -184,18 +184,18 @@ class ProductService {
...
 		product.streams.add(stream)
 		product.save(failOnError: true)
 		// A stream that is added when editing an existing free product should inherit read access for anonymous user
 		if (product.isFree()) {
-			permissionService.systemGrantAnonymousAccess(stream, Permission.Operation.READ)
+			permissionService.systemGrantAnonymousAccess(stream, Permission.Operation.STREAM_GET)
```
hpihkala: Grant both `STREAM_GET` and `STREAM_SUBSCRIBE`

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -184,18 +184,18 @@ class ProductService {
...
 		}
 		if (product.type == Product.Type.COMMUNITY) {
 			Set<SecUser> users = communityJoinRequestService.findCommunityMembers(product.beneficiaryAddress)
 			for (SecUser u : users) {
-				if (!permissionService.canWrite(u, stream)) {
-					permissionService.systemGrant(u, stream, Permission.Operation.WRITE)
+				if (!permissionService.canWriteStream(u, stream)) {
```
hpihkala: We should check `STREAM_PUBLISH` here.

As we know it's a stream, just doing something like 

`permissionService.check(u, stream, Operation.STREAM_PUBLISH)` 

avoids the need type-specific method boilerplate in `PermissionService`.

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -184,18 +184,18 @@ class ProductService {
...
+					permissionService.systemGrant(u, stream, Permission.Operation.STREAM_EDIT)
```
hpihkala: `STREAM_PUBLISH`

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -207,13 +207,13 @@ class ProductService {
 		product.save(failOnError: true)
 		// A stream that is removed when editing an existing free product should revoke read access for anonymous user
 		if (product.isFree()) {
-			permissionService.systemRevokeAnonymousAccess(stream, Permission.Operation.READ)
+			permissionService.systemRevokeAnonymousAccess(stream, Permission.Operation.STREAM_GET)
```
hpihkala: Both `STREAM_GET` and `STREAM_SUBSCRIBE`

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -207,13 +207,13 @@ class ProductService {
...
 		}
 		if (product.type == Product.Type.COMMUNITY) {
 			Set<SecUser> users = communityJoinRequestService.findCommunityMembers(product.beneficiaryAddress)
 			for (SecUser u : users) {
-				if (permissionService.canWrite(u, stream)) {
-					permissionService.systemRevoke(u, stream, Permission.Operation.WRITE)
+				if (permissionService.canWriteStream(u, stream)) {
```
hpihkala: Check `STREAM_PUBLISH`

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -207,13 +207,13 @@ class ProductService {
...
+					permissionService.systemRevoke(u, stream, Permission.Operation.STREAM_EDIT)
```
hpihkala: Revoke `STREAM_PUBLISH`

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -244,7 +244,7 @@ class ProductService {
 		product.setProperties(command.properties)
 		product.state = Product.State.DEPLOYED
 		product.save(failOnError: true)
-		permissionService.systemGrantAnonymousAccess(product)
+		permissionService.systemGrantAnonymousAccess(product, Permission.Operation.READ)
```
hpihkala: `PRODUCT_GET`

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -143,20 +143,20 @@ class StreamService {
...
-				throw new NotPermittedException(user?.username, "Stream", id, Permission.Operation.READ.id)
+				throw new NotPermittedException(user?.username, "Stream", id, Permission.Operation.STREAM_GET.id)
 			}
 		}
 	}
 
 	Set<String> getStreamEthereumPublishers(Stream stream) {
 		// This approach might be slow if there are a lot of allowed writers to the Stream
-		List<SecUser> writers = permissionService.getPermissionsTo(stream, Permission.Operation.WRITE)*.user
+		List<SecUser> writers = permissionService.getPermissionsTo(stream, Permission.Operation.STREAM_EDIT)*.user
```
hpihkala: `STREAM_PUBLISH`

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -170,12 +170,12 @@ class StreamService {
 		if (key == null || key.user == null) {
 			return false
 		}
-		return permissionService.canWrite(key.user, stream)
+		return permissionService.canWriteStream(key.user, stream)
```
hpihkala: `permissionService.check(key.user, stream, Operation.STREAM_PUBLISH)` or similar

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -170,12 +170,12 @@ class StreamService {
...
 	}
 
 	Set<String> getStreamEthereumSubscribers(Stream stream) {
 		// This approach might be slow if there are a lot of allowed readers to the Stream
-		List<SecUser> readers = permissionService.getPermissionsTo(stream, Permission.Operation.READ)*.user
+		List<SecUser> readers = permissionService.getPermissionsTo(stream, Permission.Operation.STREAM_GET)*.user
```
hpihkala: `STREAM_SUBSCRIBE`

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -189,7 +189,7 @@ class StreamService {
 		if (key == null || key.user == null) {
 			return false
 		}
-		return permissionService.canRead(key.user, stream)
+		return permissionService.canReadStream(key.user, stream)
```
hpihkala: check `STREAM_SUBSCRIBE`

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -202,17 +202,17 @@ class StreamService {
 
 	@CompileStatic
 	private boolean isDirectPermissionToStream(SecUser user, Stream stream) {
```
hpihkala: This is now too vague. Find usages and replace with direct call to `permissionService.check(user, stream, Operation.STREAM_*)` with `STREAM_*` depending on what context it is.

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -202,17 +202,17 @@ class StreamService {
...
-		return permissionService.canRead(user, stream)
+		return permissionService.canReadStream(user, stream)
 	}
 
 	@CompileStatic
 	private boolean isDirectPermissionToStream(Key key, Stream stream) {
```
hpihkala: Same as above

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -202,17 +202,17 @@ class StreamService {
...
-		return permissionService.canRead(key, stream)
+		return permissionService.canReadStream(key, stream)
 	}
 
 	@CompileStatic
 	private boolean isPermissionToStreamsUiChannelCanvas(SecUser user, Stream stream) {
-		return stream.uiChannel && stream.uiChannelCanvas && permissionService.canRead(user, stream.uiChannelCanvas)
+		return stream.uiChannel && stream.uiChannelCanvas && permissionService.canReadCanvas(user, stream.uiChannelCanvas)
```
hpihkala: The method name is vague, but I think the intent here is to `check(user, stream.uiChannelCanvas, CANVAS_GET)`

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -268,7 +268,7 @@ class StreamService {
 	def addExampleStreams(SecUser user, List<Stream> examples) {
 		for (final Stream example : examples) {
 			// Grant read permission to example stream.
-			permissionService.systemGrant(user, example, Permission.Operation.READ)
+			permissionService.systemGrant(user, example, Permission.Operation.STREAM_GET)
```
hpihkala: Grant both `STREAM_GET` and `STREAM_SUBSCRIBE`

grails-app/services/com/unifina/service/UserService.groovy
```diff
@@ -117,9 +118,9 @@ class UserService {
 
 	/** Adds/removes ModulePackage read permissions so that user's permissions match given ones */
 	def setModulePackages(user, List<ModulePackage> packages) {
```
hpihkala: I would just get rid of `ModulePackage` completely

grails-app/controllers/com/unifina/controller/api/KeyApiController.groovy
```diff
@@ -60,9 +60,15 @@ class KeyApiController {
 
 				List keys = permissionsByKey.collect { key, permissions ->
 					Map map = key.toMap()
-					map["permission"] = permissions.find {
-						it.operation == Permission.Operation.WRITE
-					}?.operation?.id ?: Permission.Operation.READ.id
+					if (Stream.isAssignableFrom(params.resourceClass)) {
+						map["permission"] = permissions.find {
+							it.operation == Permission.Operation.STREAM_EDIT
```
hpihkala: I have an idea how this whole mess could become much cleaner, but it's IMO it's not worth doing because we'll get rid of these "anonymous" (not associated with a `SecUser`) `Key`s soon anyway.

Anyway, we need to use `STREAM_PUBLISH` here.

grails-app/controllers/com/unifina/controller/api/KeyApiController.groovy
```diff
@@ -60,9 +60,15 @@ class KeyApiController {
...
+						}?.operation?.id ?: Permission.Operation.STREAM_GET.id
```
hpihkala: ...and `STREAM_SUBSCRIBE` here

grails-app/controllers/com/unifina/controller/api/KeyApiController.groovy
```diff
@@ -89,8 +95,14 @@ class KeyApiController {
 				permissionService.grant(request.apiUser, res, key, operation, false)
 
 				// If granting write, grant also read
-				if (operation == Permission.Operation.WRITE) {
-					permissionService.grant(request.apiUser, res, key, Permission.Operation.READ, false)
+				if (Stream.isAssignableFrom(params.resourceClass)) {
```
hpihkala: can assume it's a `Stream`

grails-app/controllers/com/unifina/controller/api/KeyApiController.groovy
```diff
@@ -89,8 +95,14 @@ class KeyApiController {
...
+					if (operation == Permission.Operation.STREAM_EDIT) {
```
hpihkala: Ok some magic hacks here. Well, as I said, we'll get rid of these soon. Let's try the following:

(Let's instruct the frontend guys to send the `permission` as either `stream_publish` or `stream_subscribe`)

```
// Throws if user is not permitted to grant
permissionService.grant(request.apiUser, res, key, operation, false)

// Grant extra permissions depending on whether we're creating a publish or subscribe key
if (operation == Permission.Operation.STREAM_PUBLISH) {
    // grant also STREAM_GET and STREAM_SUBSCRIBE
} else if (operation == Permission.Operation.SUBSCRIBE) {
    // grant also STREAM_GET
} else {
    // invalid request
}
```

grails-app/controllers/com/unifina/controller/api/KeyApiController.groovy
```diff
@@ -164,8 +176,8 @@ class KeyApiController {
 		}
 		Map json = new JsonSlurper().parseText((String) request.JSON)
 		String permission = request.JSON?.permission
-		if (permission && permission != "read" && permission != "write") {
-			throw new ApiException(400, "INVALID_PARAMETER", "permission field in json should be 'read' or 'write'.")
+		if (permission && permission != "stream_get" && permission != "stream_edit") {
```
hpihkala: Use the enums?

hpihkala: `stream_subscribe` and `stream_publish`

grails-app/controllers/com/unifina/controller/api/KeyApiController.groovy
```diff
@@ -174,19 +186,19 @@ class KeyApiController {
 					Permission.Operation operation = Permission.Operation.fromString(permission)
 					SecUser user = request.apiUser
 					boolean logIfDenied = false
-					if (operation == Permission.Operation.READ) {
-						if (!hasPermission(Permission.Operation.READ, k.getPermissions())) {
-							permissionService.grant(user, res, k, Permission.Operation.READ, logIfDenied)
+					if (operation == Permission.Operation.STREAM_GET) {
```
hpihkala: I guess we have to push through with this. 

`operation` from the frontend can be `STREAM_PUBLISH` or `STREAM_SUBSCRIBE`, as established above.

`if (operation == STREAM_SUBSCRIBE)` should grant (if not already granted) `STREAM_GET` and `STREAM_SUBSCRIBE`

grails-app/controllers/com/unifina/controller/api/KeyApiController.groovy
```diff
@@ -174,19 +186,19 @@ class KeyApiController {
...
 						}
-						if (hasPermission(Permission.Operation.WRITE, k.getPermissions())) {
-							permissionService.revoke(user, res, k, Permission.Operation.WRITE, logIfDenied)
+						if (hasPermission(Permission.Operation.STREAM_EDIT, k.getPermissions())) {
+							permissionService.revoke(user, res, k, Permission.Operation.STREAM_EDIT, logIfDenied)
 						}
-					} else if (operation == Permission.Operation.WRITE) {
-						if (!hasPermission(Permission.Operation.READ, k.getPermissions())) {
-							permissionService.grant(user, res, k, Permission.Operation.READ, logIfDenied)
+					} else if (operation == Permission.Operation.STREAM_EDIT) {
```
hpihkala: `if (operation == STREAM_PUBLISH)` should grant (if not already granted) `STREAM_GET`, `STREAM_SUBSCRIBE`, and `STREAM_PUBLISH`

grails-app/services/com/unifina/service/SignalPathService.groovy
```diff
@@ -287,7 +287,7 @@ class SignalPathService {
 		}
 		// All good - check if this is a stop request, which has special handling
 		else if (req.type == "stopRequest") {
-			if (!permissionService.canWrite(req.getUser(), req.getCanvas()) && !req.getUser()?.isAdmin()) {
+			if (!permissionService.canWriteCanvas(req.getUser(), req.getCanvas()) && !req.getUser()?.isAdmin()) {
```
hpihkala: check `CANVAS_STARTSTOP` permission

grails-app/controllers/com/unifina/controller/api/ModuleApiController.groovy
```diff
@@ -28,7 +28,7 @@ class ModuleApiController {
 
 	@StreamrApi
 	def index() {
-		Set<ModulePackage> allowedPackages = permissionService.getAll(ModulePackage, request.apiUser) ?: new HashSet<>()
+		Set<ModulePackage> allowedPackages = permissionService.getAll(ModulePackage, request.apiUser, Permission.Operation.DASHBOARD_GET, {}) ?: new HashSet<>()
```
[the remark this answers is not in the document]
kare: @hpihkala Do I create a migration that drops ModulePackage from db (and remove all ModulePackage related code altogether) or just remove permission service from ModuleApiController?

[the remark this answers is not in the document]
hpihkala: @kare The first option sounds better to me. It snowballs this slightly, but OTOH it also simplifies it by not having to create another set of permissions for `ModulePackages`, and it also doesn't leave technical debt.

> create a migration that drops ModulePackage from db (and remove all ModulePackage related code altogether)

kare:
@hpihkala Three failing integration tests in PermissionServiceIntegrationSpec. Marked with TODO comments. Not sure what to do with those? Remove or fix.

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -140,23 +140,25 @@ class StreamService {
...
+			if (permissionService.check(key, stream, Permission.Operation.STREAM_GET)) {
 				action.call(stream)
 			} else {
-				throw new NotPermittedException(null, "Stream", id, Permission.Operation.READ.id)
+				throw new NotPermittedException(null, "Stream", id, Permission.Operation.STREAM_GET.id)
 			}
 		} else {
-			if (isDirectPermissionToStream(user, stream) || isPermissionToStreamsUiChannelCanvas(user, stream) || isPermissionToStreamViaDashboard(user, stream)) {
+			boolean hasPermissionToStream = permissionService.check(user, stream, Permission.Operation.STREAM_GET)
+			boolean hasPermissionToStreamsUiChannelCanvas = stream.uiChannel && stream.uiChannelCanvas && permissionService.check(user, stream.uiChannelCanvas, Permission.Operation.CANVAS_GET)
```
hpihkala: This is very minor, but I noticed that in the old code, the UI channel check doesn't get executed if the user has direct permission. This saves us one call to `permissionService.check(...)`, which is expensive.

You could get the same optimization by changing the second line to
```
boolean hasPermissionToStreamsUiChannelCanvas = !hasPermissionToStream && stream.uiChannel && stream.uiChannelCanvas && permissionService.check(user, stream.uiChannelCanvas, Permission.Operation.CANVAS_GET)
```

BUT I'm not entirely sure if the new code is actually more readable than the old one? (I think it was a wise choice to delete `isDirectPermissionToStream` though)

grails-app/services/com/unifina/service/SubscriptionService.groovy
```diff
@@ -115,7 +115,7 @@ class SubscriptionService {
 		SecUser user = subscription.fetchUser()
 		if (user) {
 			streams.collect { Stream stream ->
-				permissionService.systemGrant(user, stream, Permission.Operation.READ, subscription, subscription.endsAt)
+				permissionService.systemGrant(user, stream, Permission.Operation.STREAM_SUBSCRIBE, subscription, subscription.endsAt)
```
hpihkala: I'm pretty sure we should grant both `STREAM_GET` and `STREAM_SUBSCRIBE` here

src/java/com/unifina/signalpath/AbstractSignalPathModule.java
```diff
@@ -649,7 +650,7 @@ else if (request.getType().equals("paramChange")) {
 				}
 
 				PermissionService permissionService = Holders.getApplicationContext().getBean(PermissionService.class);
-				if (!permissionService.canWrite(request.getUser(), request.getCanvas())) {
+				if (!permissionService.check(request.getUser(), request.getCanvas(), Permission.Operation.CANVAS_INTERACT)) {
```
hpihkala: For `paramChange` actually, it might make sense to require `CANVAS_EDIT`. The logic is this: If you can change a parameter when the canvas is NOT running (as permitted by `CANVAS_EDIT`), you should be allowed to change it when it IS running.

src/java/com/unifina/signalpath/ModuleWithUI.java
```diff
@@ -204,7 +205,7 @@ public void initialize() {
 
 			// User must have write permission to related Canvas in order to write to the UI channel
 			PermissionService permissionService = Holders.getApplicationContext().getBean(PermissionService.class);
-			if (!permissionService.canWrite(user, stream.getUiChannelCanvas())) {
+			if (!permissionService.check(user, stream.getUiChannelCanvas(), Permission.Operation.CANVAS_EDIT)) {
```
hpihkala: Now that `permissionService.check` encapsulates the logic for the synthetic UI channel permissions, it **should** be possible to check the "right thing" here:

```suggestion
			if (!permissionService.check(user, stream, Permission.Operation.STREAM_PUBLISH)) {
```

grails-app/services/com/unifina/service/CommunityJoinRequestService.groovy
```diff
@@ -25,11 +25,11 @@ class CommunityJoinRequestService {
 	private void onApproveJoinRequest(CommunityJoinRequest c) {
 		log.debug("onApproveJoinRequest: approved JoinRequest for address ${c.memberAddress} to community ${c.communityAddress}")
 		for (Stream s : findStreams(c)) {
-			if (permissionService.canWrite(c.user, s)) {
+			if (permissionService.check(c.user, s, Permission.Operation.STREAM_PUBLISH)) {
 				log.debug(String.format("user %s already has write permission to %s (%s), skipping grant", c.user.username, s.name, s.id))
 			} else {
 				log.debug(String.format("granting write permission to %s (%s) for %s", s.name, s.id, c.user.username))
-				permissionService.systemGrant(c.user, s, Permission.Operation.WRITE)
```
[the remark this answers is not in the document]
kare: https://streamr.atlassian.net/browse/DU-108

grails-app/domain/com/unifina/domain/data/Stream.groovy
```diff
@@ -1,13 +1,14 @@
...
-
 import com.unifina.domain.ExampleType
 import com.unifina.domain.marketplace.Product
 import com.unifina.domain.security.Permission
 import com.unifina.domain.signalpath.Canvas
 import grails.converters.JSON
 import groovy.transform.CompileStatic
+import groovy.transform.EqualsAndHashCode
 
+@EqualsAndHashCode(includes="id")
```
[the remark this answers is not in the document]
kare: @hpihkala I'm not sure about Grails, but in Java default implementation of equals() and hashCode() are inherited from java.lang.Object. Looks like Grails has the same default. http://docs.groovy-lang.org/latest/html/gapi/groovy/transform/EqualsAndHashCode.html

grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -554,16 +521,31 @@ class PermissionService {
...
+					}
+				}
+			}
 		}
 
 		return !p.empty
 	}
 
+	// map stream operations to canvas operations
+	private final static Map<Operation, Set<Operation>> STREAM_TO_CANVAS = [
```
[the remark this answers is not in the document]
kare: @hpihkala Is this line correct? https://github.com/streamr-dev/engine-and-editor/blob/CORE-1833-new-resource-permissions/grails-app/services/com/unifina/service/PermissionService.groovy#L119
Code is easy to read right now. Variables are defined close to where they are used. You can see at one glance what the mappings are. I'm not sure if it's worth the extra complexity.

grails-app/services/com/unifina/service/StreamService.groovy
kare: @hpihkala I'll optimise it performance. Readability doesn't really suffer that much.

grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -155,13 +98,30 @@ class PermissionService {
...
 		// Special case of UI channels: they inherit permissions from the associated canvas
+		Set<Permission> syntheticPermissions = new HashSet<>()
 		if (resource instanceof Stream && resource.uiChannel) {
-			directPermissions.addAll(getPermissionsTo(resource.uiChannelCanvas, userish))
+			List<Permission> permissions = getPermissionsTo(resource.uiChannelCanvas, userish)
+			for (Permission p : permissions) {
+				Set<Operation> operations = CANVAS_TO_STREAM[p.operation]
+				if (operations != null) {
+					for (Operation op : operations) {
+						syntheticPermissions.add(new Permission(stream: resource, operation: op))
```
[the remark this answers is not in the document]
kare: @hpihkala It's the Stream.id we are talking about here. All Stream's have an ID defined. Permission's ID is not listed in hashCode() or equals() implementations. I'll add a test to verify this, but I don't think this part of the code will work out of the box (without proper equals() and hashCode() implementations):
```
Set<Operation> operations = new HashSet()
for (Permission p : permissions) {
    operations.addAll(CANVAS_TO_STREAM[p.operation])
}
```

And the missing ID in Permission can cause an issue with existing permissions. I'll need to think about that.

hpihkala:
@kare some conflicts came up due to the rename PR being merged. Let me know if you need a hand resolving those!

grails-app/domain/com/unifina/domain/data/Stream.groovy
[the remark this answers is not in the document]
hpihkala: Ok

grails-app/domain/com/unifina/domain/security/Permission.groovy
```diff
@@ -5,11 +5,13 @@ import com.unifina.domain.data.Stream
...
 import com.unifina.domain.marketplace.Subscription
 import com.unifina.domain.signalpath.Canvas
-import com.unifina.domain.signalpath.ModulePackage
+import groovy.transform.CompileStatic
+import groovy.transform.EqualsAndHashCode
 
 /**
  * Access Control List (ACL) item, grants a user a specific type of access to a resource (e.g. X can read Dashboard 1)
  */
+@EqualsAndHashCode(includes="anonymous,user,key,invite,canvas,dashboard,stream,product,operation,subscription,endsAt,parent")
```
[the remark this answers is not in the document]
hpihkala: Maybe this line needs a comment explaining why it was added? It's not very future-proof: someone will forget to add some new field to the list.

grails-app/services/com/unifina/service/PermissionService.groovy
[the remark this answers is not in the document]
hpihkala: Line looks correct yes. Doing it the suggested way you wouldn't need to even ask ;) ;)

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -147,18 +147,17 @@ class StreamService {
 		}
 
 		if (key != null) {
-			if (permissionService.check(key, stream, Permission.Operation.STREAM_GET)) {
+			if (permissionService.check(key, stream, Permission.Operation.STREAM_SUBSCRIBE)) {
```
hpihkala: Why `STREAM_SUBSCRIBE` and not `STREAM_GET`?

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -147,18 +147,17 @@ class StreamService {
...
 				action.call(stream)
 			} else {
-				throw new NotPermittedException(null, "Stream", id, Permission.Operation.STREAM_GET.id)
+				throw new NotPermittedException(null, "Stream", id, Permission.Operation.STREAM_SUBSCRIBE.id)
 			}
 		} else {
-			boolean hasPermissionToStream = permissionService.check(user, stream, Permission.Operation.STREAM_GET)
```
hpihkala: Why was this row removed? Isn't this the "main" check?

rest-e2e-tests/streamr-api-clients.js
```diff
@@ -240,7 +240,7 @@ class Streams {
             .methodAndPath('POST', `streams/${id}/permissions`)
```
hpihkala: To make a stream public, need to grant (with `anonymous: true`) both `stream_get` and `stream_subscribe`

src/groovy/com/unifina/api/DashboardListParams.groovy
```diff
@@ -10,7 +10,7 @@ class DashboardListParams extends ListParams {
 
 	DashboardListParams() {
 		super()
-		operation = Permission.Operation.DASHBOARD_GET
+		operation = Permission.Operation.DASHBOARD_SHARE
```
hpihkala: Why this change? Doesn't seem like the correct default?

test/integration/com/unifina/service/PermissionServiceIntegrationSpec.groovy
```diff
@@ -308,13 +295,24 @@ class PermissionServiceIntegrationSpec extends IntegrationSpec {
...
 	}
 
+	void "getPermissionsTo(resource, userish) returns correct UI channel read permissions via associated dashboard"() {
+		service.systemGrantAll(me, dashboard)
+		def permissions = service.getPermissionsTo(uiChannelStream, me)
+
+		expect:
+		permissions.size() == 1
+		permissions.get(0).operation == Permission.Operation.CANVAS_GET
+		service.check(me, uiChannelStream, Permission.Operation.DASHBOARD_GET)
```
hpihkala: How could the user have `DASHBOARD_GET` permission to a stream?

We should check `STREAM_GET` and `STREAM_SUBSCRIBE`.

test/integration/com/unifina/service/PermissionServiceIntegrationSpec.groovy
```diff
@@ -308,13 +295,24 @@ class PermissionServiceIntegrationSpec extends IntegrationSpec {
...
 		service.check(me, stream, Permission.Operation.STREAM_DELETE)
 	}
 
+	void "getPermissionsTo(resource, userish) returns correct UI channel read permissions via associated dashboard"() {
+		service.systemGrantAll(me, dashboard)
+		def permissions = service.getPermissionsTo(uiChannelStream, me)
+
+		expect:
+		permissions.size() == 1
+		permissions.get(0).operation == Permission.Operation.CANVAS_GET
```
hpihkala: How could the user have `CANVAS_GET` permission to a stream? Can't be right. It should return 2 permissions: `STREAM_GET` and `STREAM_SUBSCRIBE`.

test/integration/com/unifina/service/PermissionServiceIntegrationSpec.groovy
```diff
@@ -328,16 +326,6 @@ class PermissionServiceIntegrationSpec extends IntegrationSpec {
 		e.message == "Cannot revoke only SHARE permission of ${stream}"
 	}
 
-	void "getPermissionsTo(resource, userish) returns correct UI channel read permissions via associated dashboard"() {
-		service.systemGrantAll(me, dashboard)
-		def permissions = service.getPermissionsTo(uiChannelStream, me)
-
-		expect:
-		permissions.size() == 1
-		permissions.get(0).operation == Permission.Operation.READ
```
hpihkala: (Related to previous comments in this file:)

Here you can see the old test, which grants access to dashboard and then checks whether the user has `READ` permission to the `uiChannelStream`. In the new permission system `READ` becomes `STREAM_GET` and `STREAM_SUBSCRIBE`.

test/unit/com/unifina/service/ApiServiceSpec.groovy
```diff
@@ -48,7 +48,7 @@ class ApiServiceSpec extends Specification {
 
 		then:
 		list.size() == 3
-		1 * permissionService.get(Dashboard, me, Permission.Operation.DASHBOARD_GET, true, _) >> [
+		1 * permissionService.get(Dashboard, me, Permission.Operation.DASHBOARD_SHARE, true, _) >> [
```
hpihkala: Why this change? Should be `DASHBOARD_GET`?

test/unit/com/unifina/service/StreamServiceSpec.groovy
```diff
@@ -160,7 +162,6 @@ class StreamServiceSpec extends Specification {
 		service.getReadAuthorizedStream("streamId", user, null, cb)
 		then:
 		thrown(NotPermittedException)
-		1 * service.permissionService.check(user, stream, Permission.Operation.STREAM_GET) >> false
```
hpihkala: This should be invoked. See earlier comment about that one (accidentally?) removed line.

test/unit/com/unifina/service/StreamServiceSpec.groovy
```diff
@@ -179,7 +180,7 @@ class StreamServiceSpec extends Specification {
 		service.getReadAuthorizedStream("streamId", null, key, cb)
 		then:
 		thrown(NotPermittedException)
-		1 * service.permissionService.check(key, stream, Permission.Operation.STREAM_GET) >> false
+		1 * service.permissionService.check(key, stream, Permission.Operation.STREAM_SUBSCRIBE) >> false
```
hpihkala: We need to think through what `getReadAuthorizedStream` is supposed to do. `STREAM_GET` seems correct, since the service method is called from the `show` action.

test/unit/com/unifina/service/StreamServiceSpec.groovy
```diff
@@ -408,10 +405,10 @@ class StreamServiceSpec extends Specification {
 		// ensuring lookup doesn't depend on case
 		boolean result2b = service.isStreamEthereumPublisher(stream, address2.toUpperCase())
 		then:
-		1 * service.permissionService.check(user1, stream, Permission.Operation.STREAM_PUBLISH) >> true
-		result1
-		1 * service.permissionService.check(user2, stream, Permission.Operation.STREAM_PUBLISH) >> false
-		!result2
+		2 * service.permissionService.check(user1, stream, Permission.Operation.STREAM_PUBLISH) >> true
```
hpihkala: Why did the call count go from 1 to 2?

grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -121,6 +144,34 @@ class PermissionService {
...
+	private Permission hasTransitiveDashboardPermissions(Canvas canvas, Userish userish) {
+		/*
+		List<DashboardItem> items = DashboardItem.findAllByCanvas(canvas)
+		if (items.isEmpty()) {
+			return null
+		}
+		*/
+		List<Permission> permissions = Permission.withCriteria() {
+			//'in'("dashboard", items.collect { it.dashboard })
+			eq("operation", Operation.CANVAS_GET)
```
hpihkala:
```suggestion
			eq("operation", Operation.DASHBOARD_GET)
```

grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -97,19 +97,42 @@ class PermissionService {
...
+			SecUser user = null
+			if (userish instanceof Key) {
+				key = userish
+			} else if (userish instanceof SecUser) {
+				user = userish
+			}
+			if (permission != null) {
+				Permission sp = new Permission(
+					canvas: resource.uiChannelCanvas,
+					operation: permission.operation,
```
hpihkala: `STREAM_GET` and `STREAM_SUBSCRIBE`

grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -97,19 +97,42 @@ class PermissionService {
 		}.toList()
 
 		// Special case of UI channels: they inherit permissions from the associated canvas
-		Set<Permission> syntheticPermissions = new HashSet<>()
 		if (resource instanceof Stream && resource.uiChannel) {
+			Set<Permission> syntheticPermissions = new HashSet<>()
+			Permission permission = hasTransitiveDashboardPermissions(resource.uiChannelCanvas, userish)
```
hpihkala: There's actually a vulnerability here. Test case:

- Create canvas
- Two modules: public (id 1) and secret (id 2). The modules have ui channel streams: module id 1 has stream `uiChannelPublic` and module 2 has stream `uiChannelSecret`.
- Create Dashboard
- Grant user `U` `DASHBOARD_GET` permission to dashboard
- Create DashboardItem with a reference to module id 1 (public).
- assert `getPermissionsTo(uiChannelPublic, U) should return `STREAM_GET` and `STREAM_SUBSCRIBE`.
- assert `getPermissionsTo(uiChannelSecret, U) should return an empty set

Maybe replace `hasTransitiveDashboardPermissions(Canvas, Userish) with

```
boolean hasPermissionViaDashboard(stream, userish)
```

grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -121,6 +144,34 @@ class PermissionService {
 		(Operation.CANVAS_STARTSTOP): [Operation.STREAM_PUBLISH],
 	]
 
+	private Permission hasTransitiveDashboardPermissions(Canvas canvas, Userish userish) {
+		/*
+		List<DashboardItem> items = DashboardItem.findAllByCanvas(canvas)
```
hpihkala: - Find out the module id by parsing `Stream#uiChannelPath`, which begins with:
`/canvases/[canvasId]/modules/[moduleId]`
- Add to this query the filter condition `eq('module', moduleId)`

grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -519,7 +570,14 @@ class PermissionService {
 		}
 
 		// Special case of UI channels: they inherit permissions from the associated canvas
-		if (p.empty && resource instanceof Stream && resource.uiChannel) {
+		if (p.isEmpty() && resource instanceof Stream && resource.uiChannel) {
+			//if (hasPermission(userish, resource.uiChannelCanvas, op)) {
```
hpihkala: Here we could make this more DRY by leveraging the `getPermissionsTo(resource, userish)` and checking if the returned `List` contains the operation `op`.

grails-app/services/com/unifina/service/StreamService.groovy
hpihkala: Maybe the whole `getReadAuthorizedStream` is now unnecessary?!

grails-app/services/com/unifina/service/PermissionService.groovy
hpihkala: Actually looks like this is pretty much implemented currently as `StreamService#isPermissionToStreamViaDashboard`. Let's refactor this to have all the transitive permission checking happening in `PermissionService`!

grails-app/domain/com/unifina/domain/data/Stream.groovy
```diff
@@ -155,4 +155,31 @@ class Stream implements Comparable {
...
+	/**
+	 * Return module id parsed from uiChannelPath or 0 if not found.
+	 */
+	@CompileStatic
+	Integer parseModuleID() {
+		if (uiChannelPath == null) {
+			return 0
+		}
+		String[] components = uiChannelPath.split("/")
+		if (components.length == 5 ) {
```
hpihkala, line 176: The `uiChannelPath` can be longer in case there are sub-canvases.

```suggestion
		if (components.length >= 5 ) {
```

grails-app/domain/com/unifina/domain/data/Stream.groovy
```diff
@@ -155,4 +155,31 @@ class Stream implements Comparable {
...
+		return uiChannel && uiChannelCanvas != null && uiChannelPath != null
+	}
+
+	/**
+	 * Return module id parsed from uiChannelPath or 0 if not found.
+	 */
+	@CompileStatic
+	Integer parseModuleID() {
+		if (uiChannelPath == null) {
+			return 0
```
hpihkala, line 173: `0` is a valid result, so the caller can't know whether the call succeeded or failed. Better to throw `NullPointerException` or `InvalidStateException` (or similar)?

grails-app/domain/com/unifina/domain/data/Stream.groovy
```diff
@@ -155,4 +155,31 @@ class Stream implements Comparable {
...
+		}
+		String[] components = uiChannelPath.split("/")
+		if (components.length == 5 ) {
+			try {
+				return Integer.parseInt(components[4])
+			} catch (NumberFormatException e) {
+				return 0
+			}
+		}
+		return 0
```
hpihkala, line 183: Also throw

grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -144,32 +151,18 @@ class PermissionService {
...
-				if (isNotNullAndIdNotNull(userish)) {
-					String userProp = getUserPropertyName(userish)
-					eq(userProp, userish)
+	private boolean isPermissionToStreamViaDashboard(Userish userish, Stream stream) {
+		if (userish != null && stream.isUIChannel()) {
+			Canvas canvas = stream.uiChannelCanvas
+			int moduleId = stream.parseModuleID()
+			List<DashboardItem> matchedItems = DashboardItem.findAllByCanvasAndModule(canvas, moduleId)
+			for (DashboardItem item : matchedItems) {
+				if (check(userish, item.dashboard, Operation.DASHBOARD_EDIT)) {
```
hpihkala, line 160: Ability to read the dashboard is enough

```suggestion
				if (check(userish, item.dashboard, Operation.DASHBOARD_GET)) {
```

grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -570,13 +563,11 @@ class PermissionService {
...
-			Permission permission = hasTransitiveDashboardPermissions(resource.uiChannelCanvas, userish)
-			if (permission != null) {
-				return true
+		if (directPermissions.isEmpty() && resource instanceof Stream && resource.uiChannel) {
+			for (Permission perm : getPermissionsTo(resource, userish)) {
+				if (perm.operation == op) {
+					return true
+				}
 			}
 			Set<Operation> operations = STREAM_TO_CANVAS[op]
```
hpihkala, line 572: Lines 572-579 are covered by new code on lines 567-570 and can be removed?

rest-e2e-tests/package.json
```diff
@@ -18,7 +18,7 @@
     "mocha": "6.2.2",
     "node-fetch": "2.6.0",
     "sleep-promise": "8.0.1",
-    "streamr-client": "3.1.3",
+    "streamr-client": "3.2.0",
```
hpihkala: Version `3.2.0` has known problems and has been deprecated from npm. Please use `3.1.3` for now

rest-e2e-tests/streamr-api-clients.js
```diff
@@ -235,7 +235,16 @@ class Streams {
             .withBody(body)
     }
 
-    makePublic(id) {
+    grantSubscribe(id) {
```
hpihkala, line 238: The method `makePublic` grants *public* permissions (=to the anonymous user), so the names `grantSubscribe` and `grantGet` don't describe what the method does very well. If you want to split the `makePublic` into atomic components, then you should either:

- Use a more accurate name like `grantPublicSubscribe` and `grantPublicGet`, or 
- Refactor them to `grantSubscribe(id, user)`, where setting `user` to `null` would grant the anonymous permission

Personally I might just keep `makePublic`... :)

test/integration/com/unifina/service/PermissionServiceIntegrationSpec.groovy
```diff
@@ -103,6 +118,57 @@ class PermissionServiceIntegrationSpec extends IntegrationSpec {
 		uiChannelStream = new Stream(name: "ui channel", uiChannel: true, uiChannelCanvas: uiChannelCanvas, uiChannelPath: "/canvases/" + uiChannelCanvas.id + "/modules/2")
 		uiChannelStream.id = "stream-id-2"
 		uiChannelStream.save(validate: true, failOnError: true)
+
+		vulCan = new Canvas(name: "canvas")
+		vulCan.save(failOnError: true, validate: true)
+		category = new ModuleCategory(name: "category")
```
hpihkala, line 124: `ModuleCategory` and `Module` are used to define the module registry, so I don't think you actually need to instantiate them here at all. The module ids in `Canvas#json`, `Stream#uiChannelPath` and `DashboardItem#module` doesn't refer to `Module#id`, instead they are just an identifier of an *instance* of a module within a canvas.

So in this test, as long as your `Stream#uiChannelPath` and `DashboardItem#module` refer to the same number (whatever number, you can still choose `1` for the allowed stream and `2` for the secret one), this test should work. So the creation of `ModuleCategory` and `Module` objects can be removed I think.

test/unit/com/unifina/service/StreamServiceSpec.groovy
```diff
@@ -382,6 +380,7 @@ class StreamServiceSpec extends Specification {
 		1 * service.permissionService.getPermissionsTo(stream, Permission.Operation.STREAM_PUBLISH) >> perms
 		addresses == validAddresses
 	}
+*/
```
hpihkala, line 383: - The test `getStreamEthereumPublishers should return Ethereum addresses of users with write permission to the Stream` should not be removed
- Remove obsolete tests instead of commenting them out

timoxley:
@kare Noticed two issues:

### Backend accepts any value for `user`

 If it's not a valid address e.g. the string "whatever", the first time you try sharing to that user, the request will fail, complaining that it couldn't send mail. 

```
Failed messages: javax.mail.SendFailedException: Invalid Addresses; nested exception is: com.sun.mail.smtp.SMTPAddressFailedException: 501 <whatever>: recipient address must contain a domain
```

On subsequent attempts, with the exact same data, it succeeds. I assume it's because it thinks it's already sent the activation email, so it doesn't try (and fail) again. API should probably either consistently fail or consistently succeed.

Also noticed there's a race condition related to this, on the initial "share", it tries to add multiple permissions to that user, and all the requests that were made in parallel will fail, but then subsequent requests will succeed. I have a feeling that the parallel requests with the "new" user will check whether an invite has been sent, and all detect that it hasn't, and thus they all try (and fail) to send the activation email, once for each permission. Should probably confirm that this can't lead to multiple identical emails being sent. I tried sharing on current production and it only sent a single email, though I do have a memory of seeing duplicate emails coming from streamr at some point in the past.

### Can add multiple copies of the same permission to the same user 

One user can end up with multiple permissions if you simply `POST` the same data twice. This is a bit problematic for permission removal because the user may still have a permission that you removed, if you didn't manage to delete all of the copies. I have worked around this in the frontend so that when you remove a permission it removes all copies of the permission (that the frontend knows about), but the backend should probably also do the same thing: if you remove a permission, remove all identical permissions for that user, not just the one referenced by the id, but in that case then the id is kinda redundant and it should just look up the permission by the user+permission type combination.

hpihkala:
@timoxley actually mentions *three* issues, out of which the **race condition** is actually the most serious one as it breaks functionality in case of sharing to a new user. It should definitely be addressed somehow.

I’m guessing it happens in cases where a new object is being created (with a uniqueness constraint). These include:

- A new `SignupInvite` in case we’re sharing to an email address not yet signed up
- A new `SecUser` in case we’re sharing to an Ethereum address not yet signed up

I think this has been previously discussed, and back then the workaround was to do permission updates sequentially instead of in parallel in the frontend. It’s still a feasible temporary fix, although not ideal.

Fixing it properly would require adding defensive code to the backend, something like:
- Try to create object
- If it fails due to uniqueness constraint violated, fetch the now-existing object and add the permissions to that instead

timoxley:
@kare I'm seeing the following error when adding new user permissions: 

> Could not find matching constructor for: com.unifina.controller.api.EmailMessage

kare:
@timoxley Yes, I'm working on it.

grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -633,8 +562,107 @@ class PermissionService {
 		}
 	}
 
-	/** null is often a valid value (but not a valid user), and means "anonymous Permissions only" */
-	private static boolean isNotNullAndIdNotNull(userish) {
-		return userish != null && userish.id != null
+	Permission savePermissionAndSendShareResourceEmail(SecUser apiUser, Key apiKey, Operation op, String targetUsername, EmailMessage msg) {
```
hpihkala, line 565: Why pass `String targetUsername` instead of `SecUser targetUser`?

grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -633,8 +562,107 @@ class PermissionService {
...
+			invite.sent = true
+			invite.save(failOnError: true, validate: true)
+		}
+		Permission newPermission = savePermission(msg.resource, apiUser, null, invite, op)
+		return newPermission
+	}
+
+	Permission savePermissionAndCreateEthereumAccount(String username, SecUser grantor, Operation operation, Resource res) {
+		EthereumIntegrationKeyService ethereumIntegrationKeyService = Holders.getApplicationContext().getBean(EthereumIntegrationKeyService)
+		SecUser user = ethereumIntegrationKeyService.createEthereumUser(username)
```
hpihkala, line 634: Repeating one earlier comment: the address might already be bound to a user, so we should `getOrCreate` the user instead of always creating it. (and change the name of this method to something like `savePermissionForEthereumAccount`.

hpihkala:
Btw notice there's some conflict with master

grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -24,102 +30,83 @@ import java.security.AccessControlException
...
-	@CompileStatic
-	void verifyShare(Userish userish, resource) throws NotPermittedException {
-		verify(userish, resource, Operation.SHARE)
+	PermissionStore store = new PermissionStore()
+	GrailsApplication grailsApplication
+	MailService mailService
+	SignupCodeService signupCodeService
+	PageRenderer groovyPageRenderer
+
+	private Object findID(Object resource) {
```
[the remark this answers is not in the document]
kare: GrailsCompileStatic annotation requires that code.

grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -581,7 +598,11 @@ class PermissionService {
 		return permission
 	}
 
+	@CompileStatic(value = TypeCheckingMode.SKIP)
 	private void sendEmailShareResource(Operation op, String username, EmailMessage msg) {
```
hpihkala, line 602: `username` argument is not used

PR #736, opened 2020-02-07
Rename communities to data unions

hpihkala:
This is a refactor that renames communities to data unions in the API and in code.

For backwards compatibility, both the old API `/communities` and the new one `/dataunions` work. The old route should be removed once it's no longer used.

Added a basic E2E test, which will fail until [this PR](https://github.com/streamr-dev/streamr-community-products/pull/39) adds the `/dataunions` route to CPS.

**Breaking API changes:** cc @juhah @tumppi 

- Field renamed on data union join request objects: `communityAddress -> contractAddress`
- Field renamed on data union secrets: `communityAddress -> contractAddress`
- When creating data union products, pass type `DATAUNION` instead of `COMMUNITY`

**Note to reviewers:** A few renamed files show up as deleted & new files, sorry. IDE+git fail.


jtakalai:
Note also https://github.com/streamr-dev/streamr-client-javascript/blob/master/src/rest/CommunityEndpoints.js

hpihkala:
> Note also https://github.com/streamr-dev/streamr-client-javascript/blob/master/src/rest/CommunityEndpoints.js

@jtakalai old URLs still work, so this shouldn't break (many) things.

kare:
@hpihkala I've got
```
[diff]
	renames = copy
	renamelimit = 3000
```
in my .gitconfig. This setting should help Git with detecting renames. There is similar option for merge too.
https://stackoverflow.com/questions/13805750/git-fails-to-detect-renaming

hpihkala:
@juhah @tumppi this renaming needs coordination with frontend. Based on a quick check, it seems you're not reading the renamed `communityAddress` variable. In that case the only breaking change is that you should pass product type `DATAUNION` instead of `COMMUNITY` when creating a data union product.

Please open a PR in `platform` and then let's coordinate merging these at the same time.

PR #740, opened 2020-02-27
ensure lower case inbox stream addresses and case-insensitive lookups

mthambipillai:


grails-app/controllers/com/unifina/controller/api/PermissionApiController.groovy
```diff
@@ -37,7 +39,12 @@ class PermissionApiController {
 		if (!resourceClass) { throw new IllegalArgumentException("Missing resource class") }
 		if (!grailsApplication.isDomainClass(resourceClass)) { throw new IllegalArgumentException("${resourceClass.simpleName} is not a domain class!") }
 
-		def res = resourceClass.get(resourceId)
+		def res
+		if (resourceClass.equals(Stream.class)) {
```
kare: This might be problematic if resourceClass is proxied by Hibernate.
`Stream.isAssignableFrom(resourceClass)` should work.

PR #741, opened 2020-03-06
CORE-1876: Add new fields for Product

kare:
- Contact
- Terms of use


grails-app/domain/com/unifina/domain/marketplace/Product.groovy
```diff
@@ -28,6 +29,85 @@ class Product {
...
+				reselling: reselling,
+				storage: storage,
+				termsUrl: termsUrl,
+				termsName: termsName,
+			]
+		}
+	}
+
+	// Product's contact details.
+	Contact contact
```
hpihkala: Should these have default values? Would that avoid having to pass them in the `Spec`s?

```
Contact contact = new Contact()
TermsOfUse termsOfUse = new TermsOfUse()
```

test/integration/com/unifina/com/unifina/domain/marketplace/PaidSubscriptionIntegrationSpec.groovy
```diff
@@ -30,6 +30,8 @@ class PaidSubscriptionIntegrationSpec extends IntegrationSpec {
 			ownerAddress: "0x0000000000000000000000000000000000000000",
 			beneficiaryAddress: "0x0000000000000000000000000000000000000000",
 			pricePerSecond: 1,
+			contact: new Product.Contact(),
```
hpihkala: This is all over the Specs now, can this kind of changes be avoided by initializing the fields with default values (see previous comment)?

kare:
I'm running into JVM crash with this PR. I don't know whether it is because of this PR (might be due to use of static inner class inside a domain class -> maybe Hibernate issue) or just surfacing some old issue. Both log file's last entry before crash is
`2020-03-25 13:42:19,706 [http-nio-8081-exec-9] INFO  service.DataUnionOperatorService  - DataUnionOperatorServiceImpl.proxy() execution took 413 ms`.
Sounds like the operation completed, but we have some issues in production with Apache connection pooling related to DU. Could it be it?
- Log 1 with a longer REST API test log and JVM dump at the end: [jvm-fatal-error.txt](https://github.com/streamr-dev/engine-and-editor/files/4382230/jvm-fatal-error.txt)
- Log 2: [jvm-fatal-error-2.txt](https://github.com/streamr-dev/engine-and-editor/files/4382236/jvm-fatal-error-2.txt)
- JVM dumps: [hs_err_replay.tar.gz](https://github.com/streamr-dev/engine-and-editor/files/4382244/hs_err_replay.tar.gz)
- Patch: [0001-Remove-new-Product.-Contact-TermOfUse-from-specs.patch.txt](https://github.com/streamr-dev/engine-and-editor/files/4382254/0001-Remove-new-Product.-Contact-TermOfUse-from-specs.patch.txt)

kare:
SIGSEGV https://travis-ci.com/github/streamr-dev/engine-and-editor/jobs/312353537

hpihkala:
@kare ok looks like there's a low-level bug somewhere, maybe in Grails or Groovy. I don't see anything else "special" except the inner classes in the domain object.

- You could try splitting off the inner classes to separate classes and see if you can still make the JVM crash. (You might want to put the classes under the `src` tree and not `grails-app/domain` so that Grails doesn't think they are new domain objects)
- If that doesn't fix the problem, maybe revert to flat field structure

kare:
@hpihkala Looks a lot better now. I ran REST tests 10 times locally and no crashes!

PR #745, opened 2020-03-30
CORE-1887: Replace where queries with criteria

kare:
There is some odd stuff going on here. I can't compile code that works elsewhere! See TODO comment(s).

Also there is a long list of `Entity.findAll` queries, but lets fix the where queries first.


grails-app/services/com/unifina/service/EthereumService.groovy
```diff
@@ -38,10 +38,11 @@ class EthereumService {
...
 	/** Checks if given user has registered the given Ethereum address in their Streamr profile */
 	boolean hasEthereumAddress(SecUser user, String ethereumAddress) {
-		// TODO: should we check that service == "ETHEREUM"?
-		IntegrationKey key = IntegrationKey.where {
-			(user == user) && (idInService == ethereumAddress)
-		}.find()
+		IntegrationKey key = IntegrationKey.createCriteria().get {
+			eq("user", user)
+			eq("idInService", ethereumAddress, [ignoreCase: true])
+			eq("service", IntegrationKey.Service.ETHEREUM)
```
kare: Is this ETHEREUM or ETHEREUM_ID?

jtakalai: ETHEREUM_ID probably. I guess ETHEREUM is they keys that are added for ethereum module integration, and ETHEREUM_ID is for user id

grails-app/services/com/unifina/service/EthereumIntegrationKeyService.groovy
```diff
@@ -121,13 +121,19 @@ class EthereumIntegrationKeyService {
 	}
 
 	SecUser getEthereumUser(String address) {
-		List<IntegrationKey> keys = IntegrationKey.findAll {
-			idInService == address && service in [IntegrationKey.Service.ETHEREUM, IntegrationKey.Service.ETHEREUM_ID]
+		IntegrationKey key = IntegrationKey.createCriteria().get {
+			'in'("service", [IntegrationKey.Service.ETHEREUM, IntegrationKey.Service.ETHEREUM_ID])
```
jtakalai: could the reason be this "in"? At least that's different. You could use `or` instead for only two cases.

Also I think maybe it should only be ETHEREUM_ID since this is used for identifying a user. I think otherwise the same key could be added to another user as integration key?

jtakalai (approved):
Good that where clause variable magic is gone

grails-app/controllers/com/unifina/controller/api/KeyApiController.groovy
```diff
@@ -116,10 +116,10 @@ class KeyApiController {
 				res.removeFromKeys(key)
 			}
 
-			def query = Permission.where {
-				key == key
+			List<Key> keys = Permission.createCriteria().list {
+				eq("key", key)
 			}
-			query.deleteAll()
+			keys*.delete(flush: true)
```
hpihkala: This does N queries, while old code does 1. Can we do this with 1 criteria query?

grails-app/services/com/unifina/service/EthereumIntegrationKeyService.groovy
kare: @jtakalai Could you try what happens on your machine when you add the third argument?

grails-app/services/com/unifina/service/EthereumIntegrationKeyService.groovy
```diff
@@ -121,13 +121,19 @@ class EthereumIntegrationKeyService {
...
+			// TODO: Code doesn't compile with third argument: "[ignoreCase: true]"
+			// TODO: groovy.lang.MissingMethodException: No signature of method: com.unifina.service.EthereumIntegrationKeyService.eq() is applicable for argument types: (java.lang.String, java.lang.String, java.util.LinkedHashMap) values: [idInService, 0xf4f683a8502b2796392bedb05dbbcc8c6e582e59, [ignoreCase:true]]
```
hpihkala: Let’s try to resolve this. Did it compile in DataUnionSecretService?

grails-app/services/com/unifina/service/EthereumIntegrationKeyService.groovy
hpihkala: Both ETHEREUM and ETHEREUM_ID

grails-app/services/com/unifina/service/EthereumService.groovy
hpihkala: Both ETHEREUM and ETHEREUM_ID.

Old code worked because those are the only two types that exist. But good idea to add them in case there are others in the future.

grails-app/services/com/unifina/service/DataUnionSecretService.groovy
```diff
@@ -22,9 +22,10 @@ class DataUnionSecretService {
 		return result
 	}
 
-	DataUnionSecret find(String contractAddress, String DataUnionSecretId) {
-		DataUnionSecret result = DataUnionSecret.where {
-			(contractAddress == contractAddress) && (id == DataUnionSecretId)
+	DataUnionSecret find(String contractAddress, String id) {
+		DataUnionSecret result = DataUnionSecret.createCriteria().get {
```
hpihkala: What does .get().find() do?

test/integration/com/unifina/service/EthereumServiceIntegrationSpec.groovy
```diff
@@ -0,0 +1,43 @@
+package com.unifina.service
+
+import com.unifina.domain.security.IntegrationKey
+import com.unifina.domain.security.SecUser
+import grails.test.spock.IntegrationSpec
+
+class EthereumServiceIntegrationSpec extends IntegrationSpec {
+	void "hasEthereumAddress query works with lowercase or uppercase"() {
+		setup:
```
hpihkala: Please add similar test for getEthereumUser, if not already present

hpihkala (changes requested):

hpihkala:
`Entity.findAllBy*` and `Entity.findBy*` queries have worked well so far, probably no need to fix them. `Entity.findAll {...}` could have issues. Are there many of the latter?

Has the Hibernate version changed in the last 6 months or so?

grails-app/services/com/unifina/service/DataUnionSecretService.groovy
kare: @hpihkala I was wondering the same but adding find() to problematic query doesn't resolve issue. I think it should be removed here.

kare:
@hpihkala There's about 20 findAll {} queries. Hibernate has not been updated.

grails-app/services/com/unifina/service/DataUnionSecretService.groovy
hpihkala: Yeah looks lie a leftover from old code. Elsewhere you're using just `.get()`.

kare:
@hpihkala @jtakalai I think the problem is with how GORM and unit testing works as explained here: https://docs.grails.org/latest/ref/Domain%20Classes/createCriteria.html

> Unit tests use an in-memory implementation of GORM without Hibernate, and as such, createCriteria() returns an instance of CriteriaBuilder, not HibernateCriteriaBuilder.

So looks like these can only be tested with integration tests.

test/integration/com/unifina/service/EthereumServiceIntegrationSpec.groovy
kare: Done

kare:
I had some false positives in the first `findAll {}` count. They all should be fixed now.

grails-app/services/com/unifina/service/DataUnionJoinRequestService.groovy
```diff
@@ -38,9 +39,14 @@ class DataUnionJoinRequestService {
 
 	protected Set<Stream> findStreams(DataUnionJoinRequest c) {
 		log.debug(String.format("entering findStreams(%s)", c))
-		List<Product> products = Product.withCriteria {
+		List<Product> products = Product.createCriteria().list {
 			eq("type", Product.Type.DATAUNION)
-			eq("beneficiaryAddress", c.contractAddress)
+			if (Environment.current == Environment.TEST) {
```
hpihkala, line 44: Hmm what's this? Shouldn't write environment-specific code. Does the `ilike` approach work all environments? If yes, why not just use that?

grails-app/services/com/unifina/service/DataUnionJoinRequestService.groovy
```diff
@@ -38,9 +39,14 @@ class DataUnionJoinRequestService {
 
 	protected Set<Stream> findStreams(DataUnionJoinRequest c) {
 		log.debug(String.format("entering findStreams(%s)", c))
-		List<Product> products = Product.withCriteria {
```
hpihkala, line 41: Why are you refactoring `withCriteria` to `createCriteria().list`? Isn't that more or less the same thing?

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -165,16 +165,19 @@ class StreamService {
...
+				'in'("id", writers*.id)
+			}
+			'in'("service", [IntegrationKey.Service.ETHEREUM, IntegrationKey.Service.ETHEREUM_ID])
 		}
 
 		return keys*.idInService*.toLowerCase() as Set
 	}
 
 	boolean isStreamEthereumPublisher(Stream stream, String ethereumAddress) {
 		IntegrationKey key = IntegrationKey.find {
```
hpihkala, line 179: Replace with criteria query?

test/integration/com/unifina/service/EthereumIntegrationKeyServiceIntegrationSpec.groovy
```diff
@@ -0,0 +1,43 @@
+package com.unifina.service
+
+import com.unifina.domain.security.IntegrationKey
+import com.unifina.domain.security.SecUser
+import spock.lang.Specification
+
+class EthereumIntegrationKeyServiceIntegrationSpec extends Specification {
+	void "getEthereumAddress query works with lowercase or uppercase input"() {
```
hpihkala, line 8:
```suggestion
	void "getEthereumUser query works with lowercase or uppercase input"() {
```

hpihkala (changes requested):
A few questions and minor notes

grails-app/services/com/unifina/service/DataUnionJoinRequestService.groovy
kare: @hpihkala `ilike` leaves a door open for SQL injection

kare: No it doesn't. It uses positional parameters inside.

grails-app/services/com/unifina/service/DataUnionJoinRequestService.groovy
kare: At the first look of the code I had trouble understanding what's the difference between:

`Product.createCriteria`

`Product.createCriteria()`

`Product.createCriteria().list`

Answer is none. So it's better to write out all the code to make it easy to read.

withCriteria is overloaded (doing two things) so it's difficult to read.

hpihkala: Ok

hpihkala (approved):
LGTM but tests are failing (e2e rest api tests). Restarted them just now in case it's flakiness.

PR #746, opened 2020-04-01
Add Access-Control-Allow-Headers=Streamr-Client to CORS responses.

timoxley:
For CORE-1885, there's a new custom HTTP header sent along with client requests containing the streamr-client version:

image

This header needs to be whitelisted otherwise the browser will prevent CORS requests that contain this header:

image

Alternatively, this PR could be changed to allow all headers in CORS requests with `'Access-Control-Allow-Headers': '*'`

Also possible this should be changed in [`CorsResourceMapper.groovy`](https://github.com/streamr-dev/engine-and-editor/blob/master/grails-app/resourceMappers/CorsResourceMapper.groovy) instead.


timoxley:
Test failure seems unrelated.

kare:
@timoxley There was some availability issues with Maven repositories yesterday. Seems to work today. You probably want to merge master since I've just updated it with some REST API test timeout and Travis Node version fixes.

I think you can just test it by adding a new REST API test like here: https://github.com/streamr-dev/engine-and-editor/tree/master/rest-e2e-tests

timoxley:
Rebased on master, added a loose test for checking CORS headers https://github.com/streamr-dev/engine-and-editor/pull/746/commits/dca4d1136f57e08d51742f1f259393e69547e89e

PR #752, opened 2020-04-14
CORE-1869: Remove /api/v1/communities mapping

kare:
Waiting for https://github.com/streamr-dev/streamr-client-javascript/pull/122
and maybe others?!


hpihkala:
Let's hold this back until no calls are being made to the old routes anymore.

kare:
`grep /communities/ /var/log/nginx/streamr_api_access.log |wc -l`

kare:
61618

hpihkala:
How many IP addresses are the requests coming from? If it's just one or two, maybe it's some old script of ours that needs updating and we can identify where it's coming from.

kare:
@hpihkala

```
# grep /communities/ /var/log/nginx/streamr_api_access.log.1 | wc -l
115771
# grep /communities/ /var/log/nginx/streamr_api_access.log.1 | cut -d ' ' -f 1 | uniq | wc -l
115617
```

hpihkala:
Ok, probably Swash is calling the old paths still. Not safe to remove yet - let's recheck after the DU 2.0 upgrade!

kare:
I wonder why new GitHub Actions aren't picked up by this branch after merge?

PR #759, opened 2020-04-15
Hotfix: Remove explicit inbox stream permission granting

hpihkala:
The inbox stream permission granting still causes problems in production by bloating the table. It looks like the code is working correctly, but this approach is fundamentally non-sustainable if there are large numbers on publishers, subscribers - or in the worst case - both.

The permission table has around 1M rows, 700k of which are inbox stream permission rows!

Screen Shot 2020-04-15 at 20 43 20

The numbers simply blow up quickly, making new grants and revokes very slow, leading to database tables being locked for extended periods, and other transactions failing. I think this approach simply won't work out, ever.

This hotfix is a temporary alleviation to the pains experienced currently in production. In addition, I'm deleting the inbox stream permission rows from the db. They can be regenerated if we ever decide to undo this change (we probably won't).

The sustainable way forward is to implement the key request and response validation as special logic instead of relying on explicit publish permissions on inbox streams. See [this ticket](https://streamr.atlassian.net/browse/CORE-1883).


hpihkala:
> LGTM; we should add a note somewhere (Jira?) to remind us of the need to re-implement permission checking logic before feature shipment.

The new plan is [here](https://streamr.atlassian.net/browse/CORE-1883). I started working on it a little bit by writing a [quick spec](https://github.com/streamr-dev/streamr-specs/blob/master/validation.md), and starting a [branch in JS protocol repo](https://github.com/streamr-dev/streamr-client-protocol-js/tree/stream-message-validator). I'll be asking to divide the work once I get a bit further down the road!

PR #762, opened 2020-04-22
Hotfix: Make StreamrClientService thread-safe

hpihkala:
Makes `StreamrClientService#getInstanceForThisEngineNode()` thread-safe


grails-app/services/com/unifina/service/StreamrClientService.groovy
```diff
@@ -8,18 +8,24 @@ import com.streamr.client.options.EncryptionOptions
 import com.streamr.client.options.SigningOptions
 import com.streamr.client.options.StreamrClientOptions
 import com.unifina.domain.security.Key
+import com.unifina.signalpath.time.ClockModule
```
harbu, line 11: Unused import slipped in?

PR #782, opened 2020-05-20
try to fix rest-e2e tests by increasing timeouts

harbu:
Tests pass locally so trying to just increase timeouts if that would be the issue.


kare:
Please don't merge just yet. I'll add one commit soon!

PR #784, opened 2020-05-22
CORE-1937: JVM crash

hpihkala:
This simple code crashes the JVM every time:

```
for (int i=0; i<2000; i++) {
	new Challenge("foo", 30, 10)
}
```

This is the constructor that crashes:

```
public Challenge(String text, int length, int ttlSeconds) {
	this.id = AlphanumericStringGenerator.getRandomAlphanumericString(length);
	this.challenge = text + id;
	this.expiration = new DateTime().plusSeconds(ttlSeconds).toDate();
}
```

However, this works, leading me to believe it's something with the `AlphanumericStringGenerator`:

```
public Challenge(String text, int length, int ttlSeconds) {
	this.id = "foo";      // AlphanumericStringGenerator.getRandomAlphanumericString(length);
	this.challenge = text + id;
	this.expiration = new DateTime().plusSeconds(ttlSeconds).toDate();
}
```

BUT this also works, so it can't be that simple.

```
for (int i=0; i<2000; i++) {
	AlphanumericStringGenerator.getRandomAlphanumericString(30)
}
```

**So I'm puzzled.**

In the process of narrowing it down, I made the following useful improvements included in this PR:

- Updated Redis library
- Updated JodaTime library
- Added `@GrailsCompileStatic` to `ChallengeService`
- Fixed thread safety issues in `Challenge#toMap()` (unsafe use of `SimpleDateFormat`)


harbu:
This reminds me of the issues I was having when upgrading this project to Java 8. Might be interesting for you to refer to the accompanying PR https://github.com/streamr-dev/engine-and-editor/pull/367. Never really understood why the JVM crashes but it seemed to be because of either (or both)
- using `@GrailsCompileStatic`
- invoking static method from Groovy (especially overloaded methods)

PR #788, opened 2020-05-29
CORE-1938: Fix migration bug

kare:


grails-app/migrations/core/2020-01-24-new-permissions.groovy
```diff
@@ -5,15 +5,17 @@ databaseChangeLog = {
 	changeSet(author: "kkn", id: "new-permissions-1") {
 		grailsChange {
 			change {
-				sql.eachRow("select id, stream_id, ends_at, parent_id, anonymous from permission where stream_id is not null and operation = 'read'") { row ->
+				sql.eachRow("select id, user_id, invite_id, stream_id, ends_at, parent_id, anonymous from permission where stream_id is not null and operation = 'read'") { row ->
```
hpihkala, line 8: Looks like the following `Permission` fields have still been overlooked:

- key
- subscription
- parent (This is tricky, because we should migrate the hierarchy of permissions. if we "just do it", it will lead to foreign key failures like we saw in staging. BUT currently the `parent` field is not used, so it should always be null. It's probably safe to ignore for now)

hpihkala: Please double-check that all fields in `Permission` are being migrated!

hpihkala:
I'll merge this to get things forward

PR #795, opened 2020-06-16
Check Content-Type of requests

hpihkala:
The API endpoints should return a proper error message if the request contains a body but the `Content-Type` header is not what the endpoint is expecting.

**TODO**

- Add tests
- Go through the endpoints and set the annotation on all endpoints that are NOT expecting JSON (image uploads, csv uploads, and such)
- Update swagger?

@kare can you take over from here?


kare:
Swagger is updated and I've tested locally uploading images (avatar and product) and CSV upload works too.

PR #796, opened 2020-06-19
[JS-6] Support inbox stream prefixes WIP

timoxley:
Part of ongoing work to integrate latest (beta) version of streamr-protocol in https://github.com/streamr-dev/streamr-client-javascript/pull/147

These changes make the client's group key tests pass, but I'm still tracking down an issue there regarding validation of publishers. 

No tests in this PR yet, WIP

### DIFF
--- a/grails-app/services/com/unifina/service/StreamService.groovy
+++ b/grails-app/services/com/unifina/service/StreamService.groovy
@@ -27,6 +27,7 @@ class StreamService {
 
 	CassandraService cassandraService
 	PermissionService permissionService
+	EthereumIntegrationKeyService ethService
 
 	private final StreamPartitioner partitioner = new StreamPartitioner()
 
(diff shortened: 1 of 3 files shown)


grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -27,6 +27,7 @@ class StreamService {
 
 	CassandraService cassandraService
 	PermissionService permissionService
+	EthereumIntegrationKeyService ethService
```
kare, line 30: There is a convention to use full name for the variable name. There also is an EthereumService class.
Rename `ethService` -> `ethereumIntegrationKeyService`

timoxley:
Looks like this is not needed as the client should never actually be querying info about streams with a `/STREAM/keyexchange/:id` prefixed id.

PR #803, opened 2020-08-17
CORE-1951: Check that the join worked before responding to join request

kare:


hpihkala:
@kare notice conflicts

grails-app/services/com/unifina/service/DataUnionJoinRequestService.groovy
```diff
@@ -47,6 +56,13 @@ class DataUnionJoinRequestService {
 			}
 		}
 		sendMessage(c, "join")
+		ThreadUtil.sleep(1000)
+		if (!isMemberActive(c.contractAddress, c.memberAddress)) {
```
hpihkala, line 60: Would be better to check every `interval` millis until a maximum of `timeout` has been reached.

Suitable params could be `interval = 1000` and `timeout = 10 * 1000`

src/groovy/com/unifina/utils/ThreadUtil.groovy
```diff
@@ -0,0 +1,19 @@
+package com.unifina.utils
+
+class ThreadUtil {
```
hpihkala, line 3: Why is this class useful? I'm not sure I understand why it was added, so maybe a comment would be useful.

hpihkala (changes requested):
Looks pretty good, but let's make the timeout longer and check regularly during that longer interval, see below

PR #805, opened 2020-08-20
Accept application/x-www-form-urlencoded

jtakalai:
Then server would accept stuff from front-end just like so far. https://streamr.network/signup probably broke after https://github.com/streamr-dev/engine-and-editor/pull/795, this PR is a quick-fix.


jtakalai:
Probably not needed since there's front-end fix https://github.com/streamr-dev/streamr-platform/pull/1083

PR #807, opened 2020-08-24
CORE-1952: Custom Stream.id ENS name

kare:


isentropy (approved):
Web3j changes look correct

PR #808, opened 2020-08-25
CORE-1947: New field "signupMethod" to "user" class

teogeb:
We set the value to "API" or "CORE" at signup. All users that exist already in the database are marked as "UNKNOWN".


grails-app/controllers/com/unifina/controller/api/AuthApiController.groovy
```diff
@@ -86,6 +92,7 @@ class AuthApiController {
 			def userProperties = [:] << cmd.properties << [
 				username: invite.email,
 				email: invite.email,
+				signupMethod: getSignupMethod()
 			]
 			user = userService.createUser(userProperties)
```
hpihkala, line 97: There are other places where users are created too! This is only used by the legacy email/password signup flow. For example, if someone signs up with Ethereum, this controller is not involved. Those other places should be covered too.

hpihkala (changes requested):

grails-app/services/com/unifina/service/StreamrClientService.groovy
```diff
@@ -76,7 +77,7 @@ class StreamrClientService {
 					// Create a custom EthereumAuthenticationMethod which doesn't call the API, but instead uses the internal services to
 					// get a sessionToken. Calling the API here can lead to a deadlock situation in some corner cases, because the
 					// service calls "itself" while blocking in a mutex-lock.
-					InternalAuthenticationMethod authenticationMethod = new InternalAuthenticationMethod(nodePrivateKey, ethereumIntegrationKeyService, sessionService)
+					InternalAuthenticationMethod authenticationMethod = new InternalAuthenticationMethod(nodePrivateKey, ethereumIntegrationKeyService, sessionService, SignupMethod.API)
```
teogeb, line 80: @hpihkala Is it ok to use SignupMethod.API here? Or should we analyze "Origin" header here, too?

hpihkala: @teogeb this is always programmatic, so let's use `API` here.

grails-app/domain/com/unifina/domain/SignupMethod.groovy
```diff
@@ -0,0 +1,20 @@
...
+
+enum SignupMethod {
+	API,
+	CORE,
+	UNKNOWN
+
+	private final static String SERVER_URL_CONFIG_KEY = "grails.serverURL"
+
+	static SignupMethod fromRequest(HttpServletRequest request) {
+		def serverUrl = MapTraversal.getString(Holders.getConfig(), SERVER_URL_CONFIG_KEY)
```
hpihkala, line 16: We always use types!

hpihkala (approved):
Minor comment re using types, otherwise LGTM!

PR #809, opened 2020-08-28
CORE-1921: updating recreates permissions

teogeb:
Sending a PUT to the /products/id endpoint seems to recreate the public Permission rows, if the product is a free product.

This PR fixes the issue, but there is a separate bug BACK-6 which will be fixed in another PR.

### DIFF
--- a/grails-app/services/com/unifina/service/PermissionService.groovy
+++ b/grails-app/services/com/unifina/service/PermissionService.groovy
@@ -69,6 +69,12 @@ class PermissionService {
 		return id != null && hasPermission(userish, resource, op)
 	}
 
+	boolean checkAnonymousAccess(Object resource, Operation op) {
+		String resourceProp = getResourcePropertyName(resource)
+		List<Permission> permissions = getPermissionsTo(resource, op)
+		return permissions.stream().filter{Permission p -> p.anonymous}.findAny().isPresent()
+	}
+
 	/**
 	 * Throws an exception if user is not allowed to perform specified operation on a resource.
 	 */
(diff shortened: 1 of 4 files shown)


grails-app/services/com/unifina/service/PermissionService.groovy
```diff
@@ -69,6 +69,12 @@ class PermissionService {
 		return id != null && hasPermission(userish, resource, op)
 	}
 
+	boolean checkAnonymousAccess(Object resource, Operation op) {
```
hpihkala, line 72: I think this is the same as `check(null, resource, op)` - you probably want to use that instead of implementing a new method, as there are some details which are now overlooked (permission expiration for example).

grails-app/services/com/unifina/service/ProductService.groovy
```diff
@@ -164,23 +165,30 @@ class ProductService {
...
-		// A stream that is added when editing an existing free product should inherit read access for anonymous user
-		// Revoke public read permissions and grant them back after update
 		Product product = findById(id, currentUser, Permission.Operation.PRODUCT_EDIT)
-		if (product.isFree()) {
-			product.streams.each { stream ->
-				permissionService.systemRevokeAnonymousAccess(stream, Permission.Operation.STREAM_GET)
-				permissionService.systemRevokeAnonymousAccess(stream, Permission.Operation.STREAM_SUBSCRIBE)
-			}
-		}
+		Set<Stream> addedStreams = command.streams.stream().filter{Stream s -> !product.streams.contains(s)}.collect(toSet()) as Set<Stream>
```
hpihkala, line 169: Minor thought: Would the Groovy way be slightly cleaner?

hpihkala (changes requested):
Ok the logic looks correct to me, only red flag is the addition of `checkAnonymousAccess` vs. using existing methods for checking permissions.

hpihkala (approved):
LGTM!

PR #812, opened 2020-09-01
BACK-3: Canvases use Ethereum-authenticated StreamrClient

kare:
Canvases use an Ethereum-authenticated StreamrClient instance to publish and subscribe.


grails-app/services/com/unifina/service/StreamrClientService.groovy
```diff
@@ -108,24 +111,30 @@ class StreamrClientService {
...
 			}
 		}
 		if (keys.isEmpty()) {
-			throw new IllegalStateException("User does not have an API key! This should not happen!")
+			User user = User.get(userId)
+			ECKeyPair keyPair = Keys.createEcKeyPair()
+			BigInteger decimal = keyPair.getPrivateKey()
+			String privateKey = decimal.toString(16)
+			IntegrationKey key = ethereumIntegrationKeyService.createEthereumAccount(user, "Auto-generated key", privateKey)
+			return key.idInService
```
kare, line 135: TODO: Return private key

grails-app/domain/com/unifina/domain/IntegrationKey.groovy
```diff
@@ -32,7 +32,9 @@ class IntegrationKey implements Serializable {
 	}
 
 	enum Service {
+		// Ethereum keys for canvases on Profile page
```
hpihkala, line 35: 👍🏻

hpihkala (approved):
Nicely done, LGTM. 

In case of multiple parallel requests there may be a race condition to autogenerating a key if one doesn't exist. However:

- it shouldn't be very common
- worst that can happen is that multiple keys get generated
- it will only happen max once for a user, since one they have existing keys those will be used

It's probably ok to ignore the above for now, although we've seen that this kind of race conditions are easily created. Would be good to have a standard way of solving them (different locking strategy?). 🤷🏻

PR #817, opened 2020-09-08
BACK-21: Move REST API (UI Layer) classes to single package com.unifina.controller

kare:
Remove dependencies from service to controller.


teogeb:
Some command objects are now in `controller` package and some in `service` package. Maybe we could move all commands to `service` package? Or create a separate `command` package?

teogeb (approved):

PR #823, opened 2020-09-10
BACK-9: Inherit Authorization header when proxying SignalPath runtime requests

teogeb:
Gets the Authorization header from the HTTP requests and passes that header to ApiService.post. Replaces deprecated API key authorization functionality.

Removed also SignalPathService.ping() method which was not used from controller.


kare (approved):
LGTM

PR #825, opened 2020-09-11
BACK-6: Anonymous permissions should not be revoked when a stream belongs multiple free products

teogeb:
Fixes a bug where a stream is removed from a product, but it still belongs to another free product. The stream should retain the anonymous access permission.


kare (approved):
LGTM

PR #830, opened 2020-09-16
BACK-31: Add Data Union version field to Product

teogeb:
Added new "dataUnionVersion" field to product and updated POST /api/v1/products to read that from input JSON.


kare (approved):
LGTM

PR #835, opened 2020-09-16
BACK-40: Rename DataUnionOperator -> DataUnion

kare:


teogeb (changes requested):
We could wait until we drop support for DU v1 and clean up the code base then.

kare:
Name is just wrong. Term Operator is unnecessary.

teogeb (approved):

PR #836, opened 2020-09-16
BACK-38: Separate stress tests

teogeb:
Separate normal end-to-end tests and stress tests so that we can run either of those test suites easily.

There are three NPM scripts:
- ```npm test``` - all tests
- ```npm test:e2e``` - normal end-to-end tests
- ```npm test:stress``` - stress tests


kare (approved):
LGTM

PR #840, opened 2020-09-22
BACK-12: Convert api keys to Ethereum IDs

teogeb:
All API keys are converted to Ethereum integration keys. 

The database migration script converts API keys to integration keys with a deterministic function, inserts the keys to the DB and deletes the API keys. It creates new anonymous users for keys that don't belong to any user.

API key authentication is still supported: when a user authenticates with an API key, TokenAuthenticator calculates the integration key from the API key and finds the user by that integration key.

After this PR has been merged, we can most likely remove all remaining API key functionality (expect the TokenAuthenticator functionality described above) and drop the `Key` from the DB. Those changes will be done in a separate PR.

Open questions:

- There is an endpoint `/api/v1/login/apikey` where a user can create a session token by posting an API key. Is that endpoint still supported? Should we convert the API key to an integration key there similarly as we do in TokenAuthenticator? (`LoginApiController.groovy:49`)

- ApiKeyConverter uses PBKDF2WithHmacSHA512 to create an integration key. That algorithm requires a salt. Is it ok to use minimalistic 1-byte salt value that we define in the source code? Or should we define application level configuration property and read the salt from there? Currently we use just value "0" as a salt (`ApiKeyConverter.java:13`). Spring Security's [Pbkdf2PasswordEncoder](https://github.com/spring-projects/spring-security/blob/master/crypto/src/main/java/org/springframework/security/crypto/password/Pbkdf2PasswordEncoder.java) implements PBKDF2 encoding by generating 8 random bytes as salt and prefixing the salt to the output in plain text. We don't include the salt in the output. Is that ok?

- When we migrate an API key that belongs to a user, we create a new integration key and delete all permissions that are attached to the key (`"DELETE FROM permission WHERE key_id = ?", [apiKeyId]`). Is it ok just to delete all permissions, and not migrate those? In the test DB there are no permissions for these keys, but the production environment may have some? (`2020-09-18-convert-API-keys-to-Ethereum-IDs.groovy:87`)

- When we create an API key that doesn't belong to a user, we create a new user. Is the "UNKNOWN" signupMethod a good value to be used for these new users? Or should we e.g. create a new enum value "MIGRATED" for these? (`2020-09-18-convert-API-keys-to-Ethereum-IDs.groovy:113`)

### DIFF
--- a/grails-app/migrations/core/2020-09-18-convert-API-keys-to-Ethereum-IDs.groovy
+++ b/grails-app/migrations/core/2020-09-18-convert-API-keys-to-Ethereum-IDs.groovy
@@ -185,73 +185,13 @@ databaseChangeLog = {
 		}
 	}
 
-	// create inbox streams for the integration keys
-	changeSet(author: "teogeb", id: "convert-API-keys-to-Ethereum-IDs-5") {
-		grailsChange {
-			change {
-				sql.execute("""
-					INSERT INTO stream (version, id, name, date_created, last_updated, partitions, require_signed_data, auto_configure, storage_days, inbox, inactivity_threshold_hours, example_type, require_encrypted_data)
-					SELECT
-						0,
-						account_address,
-						account_address,
-						CURRENT_TIMESTAMP(),
-						CURRENT_TIMESTAMP(),
-						1,
-						false,
-						false,
-						?,
-						true,
-						?,
-						?,
-						false
-					FROM tmp_apikey_lookup
-				""", [Stream.DEFAULT_STORAGE_DAYS, Stream.DEFAULT_INACTIVITY_THRESHOLD_HOURS, ExampleType.NOT_SET.value])
-			}
-		}
-	}
-
-	// create permissions for the inbox streams
-	changeSet(author: "teogeb", id: "convert-API-keys-to-Ethereum-IDs-6") {
-		grailsChange {
-			change {
-				List<Permission.Operation> permissionOperations = [
-					Permission.Operation.STREAM_GET,
-					Permission.Operation.STREAM_EDIT,
-					Permission.Operation.STREAM_DELETE,
-					Permission.Operation.STREAM_PUBLISH,
-					Permission.Operation.STREAM_SUBSCRIBE,
-					Permission.Operation.STREAM_SHARE,
-				];
-				sql.execute("""
-					INSERT INTO permission (version, operation, stream_id, anonymous, user_id)
-					SELECT
-						0,
-						permission_id,
-						account_address,
-						false,
-						user_id
-					FROM tmp_apikey_lookup
-					JOIN (
-						(SELECT ? as permission_id FROM DUAL) UNION ALL
-						(SELECT ? as permission_id FROM DUAL) UNION ALL
-						(SELECT ? as permission_id FROM DUAL) UNION ALL
-						(SELECT ? as permission_id FROM DUAL) UNION ALL
-						(SELECT ? as permission_id FROM DUAL) UNION ALL
-						(SELECT ? as permission_id FROM DUAL)
-					) as permission_ids
-				""", permissionOperations.collect { it.id })
-			}
-		}
-	}
-
 	// delete keys
-	changeSet(author: "teogeb", id: "convert-API-keys-to-Ethereum-IDs-7") {
+	changeSet(author: "teogeb", id: "convert-API-keys-to-Ethereum-IDs-5") {
 		sql('DELETE FROM `key`')
 	}
 
 	// drop lookup table
-	changeSet(author: "teogeb", id: "convert-API-keys-to-Ethereum-IDs-8") {
+	changeSet(author: "teogeb", id: "convert-API-keys-to-Ethereum-IDs-6") {
 		dropTable(tableName: "tmp_apikey_lookup")
 	}
 }
--- a/grails-app/migrations/core/2020-09-18-convert-API-keys-to-Ethereum-IDs.groovy
+++ b/grails-app/migrations/core/2020-09-18-convert-API-keys-to-Ethereum-IDs.groovy
@@ -1,0 +1,83 @@
+package core
+
+import com.unifina.domain.IntegrationKey
+import com.unifina.utils.IdGenerator
+import com.unifina.security.ApiKeyConverter
+import com.unifina.service.EthereumIntegrationKeyService
+import com.unifina.utils.AlphanumericStringGenerator
+import grails.converters.JSON
+
+def getAccountAddress(String apiKeyId) {
+	String privateKey = ApiKeyConverter.createEthereumPrivateKey(apiKeyId)
+	return EthereumIntegrationKeyService.getAddress(privateKey)
+}
+
+databaseChangeLog = {
+
+	// For each API key that is attached to a user:
+	// - create a new IntegrationKey for the user
+	changeSet(author: "teogeb", id: "convert-API-keys-to-Ethereum-IDs-1") {
+		grailsChange {
+			change {
+				sql.eachRow('SELECT id, name, user_id FROM `key` where user_id is not null') { row ->
+					String apiKeyId = row['id']
+					String apiKeyName = row['name']
+					String userId = row['user_id']
+					String accountAddress = getAccountAddress(apiKeyId)
+					String integrationKeyName = 'Converted from API key: ' + apiKeyName
+					Date now = new Date()
+					sql.executeInsert("INSERT INTO integration_key (id, version, name, json, user_id, service, id_in_service, date_created, last_updated) VALUE (?, ?, ?, ?, ?, ?, ?, ?, ?)", [
+						IdGenerator.get(),
+						0,
+						integrationKeyName,
+						([ address: accountAddress ] as JSON).toString(),
+						userId,
+						IntegrationKey.Service.ETHEREUM_ID.name(),
+						accountAddress,
+						now,
+						now
+					])
+				}
+			}
+		}
+	}
+
+	// For each anonymous API key:
+	// - create a new Ethereum user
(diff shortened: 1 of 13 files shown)


hpihkala:
@teogeb I didn't review any code yet, just answering the open questions:

> * There is an endpoint `/api/v1/login/apikey` where a user can create a session token by posting an API key. Is that endpoint still supported? Should we convert the API key to an integration key there similarly as we do in TokenAuthenticator? (`LoginApiController.groovy:49`)

Yes, let's still keep this endpoint and use the same conversion. So this step is intended to be a non-breaking change, while we already take a major step towards unifying things under the hood by internally only using Ethereum keys. Then at some later time we drop the `/login/apikey` endpoint.

> * ApiKeyConverter uses PBKDF2WithHmacSHA512 to create an integration key. That algorithm requires a salt. Is it ok to use minimalistic 1-byte salt value that we define in the source code? Or should we define application level configuration property and read the salt from there? Currently we use just value "0" as a salt (`ApiKeyConverter.java:13`). Spring Security's [Pbkdf2PasswordEncoder](https://github.com/spring-projects/spring-security/blob/master/crypto/src/main/java/org/springframework/security/crypto/password/Pbkdf2PasswordEncoder.java) implements PBKDF2 encoding by generating 8 random bytes as salt and prefixing the salt to the output in plain text. We don't include the salt in the output. Is that ok?

We don't need a salt here because the API keys are randomly generated, so there's no danger of commonly used inputs like there is for human-selected passwords. Not only that, but we actually _should_ make the derivation function public, so that 1) users can derive their private key from their API key independently at any later time if necessary, and 2) if we want, we can implement the same key derivation function in `broker` to sign data published with an API key (this will of course later be dropped as well).

So long story short, we can use any salt, but we should disclose what value was used (and hard-coding it in the codebase is one way to make it publicly known).

> * When we migrate an API key that belongs to a user, we create a new integration key and delete all permissions that are attached to the key (`"DELETE FROM permission WHERE key_id = ?", [apiKeyId]`). Is it ok just to delete all permissions, and not migrate those? In the test DB there are no permissions for these keys, but the production environment may have some? (`2020-09-18-convert-API-keys-to-Ethereum-IDs.groovy:87`)

Permissions are attached to a key or a user, but never both. So if you're migrating a `Key` that belongs to a `user`, the `DELETE` statement above will do nothing. (You still need this statement when migrating a `Key` where `user == null`)

> * When we create an API key that doesn't belong to a user, we create a new user. Is the "UNKNOWN" signupMethod a good value to be used for these new users? Or should we e.g. create a new enum value "MIGRATED" for these? (`2020-09-18-convert-API-keys-to-Ethereum-IDs.groovy:113`)

Yes good idea, let's create a new enum value `MIGRATED`. We can use it at some later time to identify these users and do some cleanup.

teogeb:
@hpihkala All open questions are now resolved and this PR is ready for a review.

hpihkala:
Still didn't review yet, but one thing came to mind:

How long does the key derivation take? There can be quite a lot of API requests coming in and requiring the key derivation, so if it's an expensive computation, I'm wondering if some kind of in-memory caching is required.

teogeb:
> Still didn't review yet, but one thing came to mind:
> 
> How long does the key derivation take? There can be quite a lot of API requests coming in and requiring the key derivation, so if it's an expensive computation, I'm wondering if some kind of in-memory caching is required.

@hpihkala One derivation of apiKey->integrationKeyAddress takes about in 10ms in local dev environment (tested by running 5 test runs on 1000 derivations, each test run took 9486-10640 ms).

@kare benchmarked also the migration script performance. In development environment the migration of all production DB data took 13 minutes for API keys that are attached to a users, and 50 minutes for anonymous API keys.

hpihkala:
@teogeb okay thanks, that's really useful information. Sounds like there's no need for caching, 10ms is fast enough. We may need to think about the migration a bit though, we don't want the backend to be down for an hour or more while the migration is running. A few thoughts:
- Can the migration be parallelized to make it finish in 5-10 minutes? 
- Can we precompute the migration on a database copy, then insert the new rows to the production database and clean up the deleted rows?

Any other ideas?

grails-app/migrations/core/2020-09-18-convert-API-keys-to-Ethereum-IDs.groovy
```diff
@@ -0,0 +1,122 @@
...
+		([ address: accountAddress ] as JSON).toString(),
+		userId,
+		IntegrationKey.Service.ETHEREUM_ID.name(),
+		accountAddress,
+		now,
+		now
+	]);
+}
+
+def createInboxStream(long userId, String accountAddress, Sql sql) {
```
hpihkala, line 35: Hmm. Where did this inbox stream thing come from? It doesn't exist anymore, shouldn't be needed.

grails-app/migrations/core/2020-09-18-convert-API-keys-to-Ethereum-IDs.groovy
```diff
@@ -0,0 +1,122 @@
...
+					String integrationKeyName = 'Converted from API key: ' + apiKeyName
+					createIntegrationKey(integrationKeyName, accountAddress, userId, sql)
+				}
+			}
+		}
+		sql('DELETE FROM `key` where user_id is not null')
+	}
+
+	// For each anonymous API key:
+	// - create a new Ethereum user (and inbox stream for that user)
```
hpihkala, line 93:
```suggestion
	// - create a new Ethereum user
```

grails-app/migrations/core/2020-09-18-convert-API-keys-to-Ethereum-IDs.groovy
```diff
@@ -0,0 +1,122 @@
...
+						"Anonymous User",
+						AlphanumericStringGenerator.getRandomAlphanumericString(32),
+						false,
+						accountAddress,
+						SignupMethod.MIGRATED.name()
+					]);
+					int userId = insertResult[0][0]
+					sql.execute("UPDATE permission SET user_id = ?, key_id = null WHERE key_id = ?", [ userId, apiKeyId])
+					createIntegrationKey(accountAddress, accountAddress, userId, sql)
+					createInboxStream(userId, accountAddress, sql)
```
hpihkala, line 116:
```suggestion
```

grails-app/migrations/core/2020-09-18-convert-API-keys-to-Ethereum-IDs.groovy
```diff
@@ -0,0 +1,122 @@
...
+						false,
+						true,
+						"Anonymous User",
+						AlphanumericStringGenerator.getRandomAlphanumericString(32),
+						false,
+						accountAddress,
+						SignupMethod.MIGRATED.name()
+					]);
+					int userId = insertResult[0][0]
+					sql.execute("UPDATE permission SET user_id = ?, key_id = null WHERE key_id = ?", [ userId, apiKeyId])
```
hpihkala, line 114: Here or in a separate migration that happens before this one, you might want to fix the issue with anonymous keys having `stream_edit` and `stream_delete` permissions

grails-app/migrations/core/2020-09-18-convert-API-keys-to-Ethereum-IDs.groovy
```diff
@@ -0,0 +1,122 @@
...
+					createIntegrationKey(accountAddress, accountAddress, userId, sql)
+					createInboxStream(userId, accountAddress, sql)
+				}
+			}
+		}
+		sql('DELETE FROM `key` where user_id is null')
+	}
```
hpihkala, line 121: Can we now drop the `key` table and the `Key` domain object?

grails-app/services/com/unifina/service/UserService.groovy
```diff
@@ -183,4 +183,8 @@ class UserService {
 			return null
 		}
 	}
+
+	User getUserByUsername(String username) {
```
hpihkala, line 187: This method doesn't seem to be used anywhere? (Also we shouldn't ever need to look up users by username)

hpihkala (changes requested):
Code looks really good already, almost all comments are about the migration.

Additionally: any updates on increasing the migration speed? How long does it take on the prod dump on your laptop?

grails-app/migrations/core/2020-09-18-convert-API-keys-to-Ethereum-IDs.groovy
teogeb: Ok, we'll do BACK-61 before this one.

grails-app/migrations/core/2020-09-18-convert-API-keys-to-Ethereum-IDs.groovy
teogeb: There are still a lot of places where keys are used. BACK-57 and BACK-58 handle this.

grails-app/services/com/unifina/service/UserService.groovy
teogeb: Removed the method.

grails-app/migrations/core/2020-09-18-convert-API-keys-to-Ethereum-IDs.groovy
teogeb: Discussed earlier with Eric and understood from that discussion that it would be best to add inbox streams. But if they are not needed anymore, I'll remove this functionality here.

teogeb:
The optimized script uses parallelization and less SQL query invocations. It tooks approximately 8 minutes to migrate all production keys (90000 items) in local development environment.

kare (approved):
LGTM

hpihkala (approved):
Very clever! LGTM, notice conflicts.

PR #841, opened 2020-09-24
BACK-58: Remove API key endpoints

teogeb:
Remove endpoints that list/create/update/delete API keys. 

There are endpoints for user keys and stream keys:
```
GET, POST, PUT, DELETE /api/v1/users/me/keys
GET, POST, PUT, DELETE /api/v1/streams/:id/keys
```


hpihkala (approved):
Looks ok, but note that this can't be merged before the conversion PR is merged.

We could also drop the `Key` domain object and `key` table in this PR.

kare (approved):
LGTM

PR #847, opened 2020-10-01
BACK-61: Anonymous keys don't need "stream_delete" and "stream_edit" permissions

teogeb:
Delete all "stream_delete" and "stream_edit" permissions that are attached to anonymous keys.

An earlier migration script `2020-01-24-new-permissions.groovy` created most/all of these permissions when it migrated old "write" permissions to "stream_edit"+"stream_publish"+"stream_delete".


kare (approved):
LGTM

PR #848, opened 2020-10-01
BACK-63: Remove inbox stream functionality

teogeb:
Remove implementation of inbox stream functionality. Delete inbox streams from DB, permissions of those streams and inbox-column of `stream` table.

The migration scripts deletes ~22k inbox streams and ~132k permissions.


kare (approved):
LGTM

PR #850, opened 2020-10-07
BACK-67: Create Stream ID with Sandbox name

kare:
Allow users to define explicit ID when they create a new stream. The stream ID must start with "sandbox" and have a some path appended to that. E.g. "sandbox/foo/bar" is a valid stream name. 

Stream id is not a required field: if we don't specify an ID, the backend generates a random ID.


hpihkala:
Some fixes related to stream id encoding + e2e tests available in #852 based on this branch

hpihkala:
Before merging, make the following are updated:
- ~dev nginx~
- ~dev tomcat~
- staging nginx
- staging tomcat
- prod nginx
- prod tomcat

grails-app/services/com/unifina/service/StreamService.groovy
```diff
@@ -38,27 +49,42 @@ class StreamService {
...
+			partitions: cmd.partitions,
+			uiChannel: cmd.uiChannel,
+			requireSignedData: cmd.requireSignedData,
+			requireEncryptedData: cmd.requireEncryptedData,
+			uiChannelPath: uiChannelPath,
+			uiChannelCanvas: uiChannelCanvas
+		)
+		if (cmd.id != null) {
+			if (validateIDField && !CustomStreamIDValidator.validate(cmd.id)) {
+				throw new ValidationException(new FieldError("stream", "id", null))
```
harbu, line 70: Is it possible to attach an human-readable explanation / message to the `ValidationException`  explaining what went wrong? When the Exception gets eventually rendered as a JSON message, it could be useful for the end-user to know what went wrong.

rest-e2e-tests/streams-api.test.js
```diff
@@ -25,6 +25,108 @@ describe('Streams API', () => {
         streamId = response.id
     })
 
+	describe('POST /api/v1/streams', function() {
+
+		it('happy path', async function() {
+			const assertValidResponse = async (response, json, properties, expectedId) => {
```
harbu, line 31: No need for `async`?

harbu (approved):

grails-app/services/com/unifina/service/StreamService.groovy
teogeb: The error messages in the JSON:
`{
  "code": "VALIDATION_ERROR",
  "message": "Validation failed for fields:\nid (null)\n"
}`

So it says that the problem is in ID field, and something in the format of that field. We validate the ID with a regex, and therefore we don't know how the regex doesn't match. @harbu Would you like to get very detailed info like "Stream id contains an invalid character ~"?

rest-e2e-tests/streams-api.test.js
teogeb: Fixed

src/groovy/com/unifina/domain/CustomStreamIDValidator.groovy
```diff
@@ -0,0 +1,23 @@
+package com.unifina.domain
+
+import grails.compiler.GrailsCompileStatic
+
+@GrailsCompileStatic
+class CustomStreamIDValidator {
+
+	public static final String SANDBOX_PREFIX = "sandbox"
+	public static final String PATH_REGEX = "^/(?:[\\w\\.-]+/?)*\\w\$"
```
hpihkala, line 9: Could use an explanatory comment for better readabiity. In natural language, what are the rules?

src/groovy/com/unifina/domain/CustomStreamIDValidator.groovy
```diff
@@ -0,0 +1,23 @@
+package com.unifina.domain
+
+import grails.compiler.GrailsCompileStatic
+
+@GrailsCompileStatic
+class CustomStreamIDValidator {
```
hpihkala, line 6: We tend to prefer Java classes unless there's a reason to use Groovy (Grails artifacts etc., or some particular thing is much easier to write). In this class it doesn't seem to add value IMO.

src/groovy/com/unifina/utils/JSONUtil.groovy
```diff
@@ -0,0 +1,17 @@
+package com.unifina.utils
+
+import com.google.gson.Gson
+import com.google.gson.GsonBuilder
+
+import java.text.DateFormat
+
+class JSONUtil {
```
hpihkala, line 8: Prefer Java if no reason to use Groovy

src/java/com/unifina/signalpath/streams/CreateStream.java
```diff
@@ -58,14 +60,14 @@ public void clearState() {
 		cachedStreamIdsByName.clear();
 	}
 
-	private Map<String, Object> buildParams() {
-		Map<String, Object> params = new HashMap<>();
-		params.put("name", nameInput.getValue());
-		params.put("description", description.getValue());
+	private CreateStreamCommand buildCommand() {
```
hpihkala, line 63: This module should now have a way to define the `id`. It's pretty easy to add:
```
// In declarations
private final StringInput idInput = new StringInput(this, "id");

// in init()
addInput(idInput);

// in buildCommand() - only set id if non-empty
if (!idInput.getValue().isEmpty()) {
    cmd.setId(idInput.getValue());
}
```

test/unit/com/unifina/domain/CustomStreamIDValidatorSpec.groovy
```diff
@@ -0,0 +1,23 @@
...
+		value | expected
+		"sandbox/a" | true
+		"sandbox/abc/def" | true
+		"sandbox/abc/def/file.txt" | true
+		"sandbox/foo.bar/lorem.ipsum" | true
+		"sandbox/foo-bar/lorem.ipsum" | true
+		"sandbox/abc/def/" | false
+		"sandbox/abc/def/file." | false
+		"sandbox/foo//bar" | false
+		"foobar.eth/abc/def" | false
```
hpihkala, line 21: Maybe add some tests with illegal characters present

test/unit/com/unifina/service/StreamServiceSpec.groovy
```diff
@@ -66,15 +66,15 @@ class StreamServiceSpec extends Specification {
 
 	void "createStream replaces empty name with default value"() {
 		when:
-		Stream s = service.createStream([name: ""], me)
+		Stream s = service.createStream(new CreateStreamCommand(name: ""), me)
```
hpihkala, line 69: I don't think `name` is required anymore, could be cleaned up

test/unit/com/unifina/domain/CustomStreamIDValidatorSpec.groovy
hpihkala, line 21: A few more for good measure:
- `"foo"` should fail
- `"sandbox"` should fail
- `""` should fail
- `" "` should fail

hpihkala (changes requested):
Nice! Ran the branch and it works. All comments are pretty minor ones.

grails-app/services/com/unifina/service/StreamService.groovy
teogeb: Created a separate ticket for this: BACK-78

hpihkala (approved):
Perfect!

PR #853, opened 2020-10-21
BACK-75: End-to-end tests: encode URL path parameters

teogeb:
End-to-end tests should encode URL parameters to support e.g. custom stream IDs.


hpihkala (approved):
LGTM but tests failing due to:

```
  1) Permissions API
       "before all" hook in "Permissions API":
     Error: Request to http://localhost/api/v1//login/challenge/0x8875d355a9285F594dCB0ae72eC8771C69152DA9 returned with error code 404.
```

Note the double slash. This is fixed in #850.

PR #854, opened 2020-10-21
BACK-76: End-to-end tests: StreamrClient API url config

teogeb:
The commit #00496f in streamr-docker-dev commit changed the functionality of Nginx: restUrl parameter doesn't work if it has an slash at the end of the string.


hpihkala (approved):
Also in #850

PR #856, opened 2020-10-26
BACK-79: Remove API-key permission handling

teogeb:
API keys are deprecated and all keys have been migrated to Ethereum integration keys. There are no permissions that belong to keys anymore.

Remove the functionality that handles key permissions and remove "key_id" column from permission table.


kare (approved):
LGTM

PR #863, opened 2020-10-30
BACK-77: Convert e2e-tests to use private key authentication

teogeb:
Converted most of the endpoints to use private key authentication. API authentication is used only in login endpoints (for testing API key related functionality). 

The tests create new users when possible so that there is no dependency to the DB state. 

Also converted the tests to use the official JS client where it supports the required endpoint.

Most of the tests users in the DB could be removed if not used outside of E&E. The only user that the tests need is "devOpsUser" (id=6).


kare (approved):
LGTM

PR #865, opened 2020-11-05
BACK-7: Storage node case sensitivity

teogeb:
Support case-insensitive input of storage node address. Store the address to DB in Ethereum checksum form and output the JSONs in that form.

EthereumAddress object is parsed from the command JSON manually and StreamStorageNode's field type is just a plain string instead of an EthereumAddress instance. If it is possible, string<->EthereumAddress conversions should happen automatically. We can enhance the functionality in BACK-85 if we apply this logic to the whole application.

Another small functionality change: if the request path has a malformed Ethereum address, the endpoint returns HTTP 422 (previously  HTTP 400).

### DIFF
--- a/docs/swagger.json
+++ b/docs/swagger.json
@@ -2890,6 +2890,9 @@
 				},
 				"500": {
 					"description": "Unexpected error"
+				},
+				"422": {
+					"description": "Validation error"
 				}
 			}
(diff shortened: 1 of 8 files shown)


kare (approved):
LGTM

PR #868, opened 2020-11-10
BACK-86: Fix streamr-client-testing

kare:
- Run `stream-cleartext-signed` client tests instead of `stream-encrypted-shared-rotating-signed`
- Other client test options fail locally
- Expecting step to be green from now on
- Step is still allowed to fail


hpihkala (approved):
Ok, we just need to remember to revisit this once the clients can pass all the test scenarios. The failing test (with allowed to fail) has been a visible reminder.

PR #873, opened 2020-11-18
BACK-97: Drop support for "sandbox" domain

teogeb:
Remove the hard-coded shared `sandbox` domain, as it is now redundant after BACK-94.

### DIFF
--- a/test/unit/com/unifina/domain/CustomStreamIDValidatorSpec.groovy
+++ b/test/unit/com/unifina/domain/CustomStreamIDValidatorSpec.groovy
@@ -5,33 +5,34 @@ import spock.lang.Unroll
 
 class CustomStreamIDValidatorSpec extends Specification {
 
-	private static final String MOCK_DOMAIN = "sub.my-domain.eth"
+	private static final String MOCK_ENS_DOMAIN = "sub.my-domain.eth"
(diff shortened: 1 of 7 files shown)


test/unit/com/unifina/domain/CustomStreamIDValidatorSpec.groovy
```diff
@@ -5,33 +5,34 @@ import spock.lang.Unroll
...
+		"0xaBCDEABcde123456789012345678901234567890/abc/def" | true
+		"sub.my-domain.eth/foo-bar/lorem~ipsum" | false
+		"sub.my-domain.eth/abc/def/" | false
+		"sub.my-domain.eth/abc/def/file." | false
+		"sub.my-domain.eth/foo//bar" | false
 		"foobar.eth/abc/def" | false
+		"sub.my-domain.eth" | false
 		"0x1111111111111111111111111111111111111111/abc/def" | false
 		"foo" | false
-		"sandbox" | false
```
kare, line 34: Do we need a test for "0x1111111111111111111111111111111111111111"?

teogeb: No, because we already test `0x1111111111111111111111111111111111111111/abc/def` (where user doesn't own the address) and `sub.my-domain.eth` (where user owns the address, but the syntax is malformed).

kare (commented):
LGTM

kare (approved):

PR #876, opened 2020-11-24
Developer Tools Update 2020/11

kare:
- Rename test-rest to test-e2e
- Add recipes shell, ps and update
- Move e2e testing recipes to rest-e2e-tests/Makefile

### DIFF
--- a/rest-e2e-tests/Makefile
+++ b/rest-e2e-tests/Makefile
@@ -1,0 +1,78 @@
+#
+# https://makefiletutorial.com
+# https://clarkgrubb.com/makefile-style-guide
+#
+
+# This Makefile uses .ONESHELL option supported by Make 3.82
+ifeq ($(filter oneshell,$(.FEATURES)),)
+$(error error: Your version of make ($(shell make -v|head -1|cut -f 3 -d ' ')) does not support .ONESHELL)
+endif
+
+LANG := en_US.UTF-8
+SHELL := bash
+.SHELLFLAGS := -eu -o pipefail -c # run '/bin/bash ... -c /bin/cmd'
+.ONESHELL:
+.DEFAULT_GOAL := all
+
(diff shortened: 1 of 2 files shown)


rest-e2e-tests/Makefile
```diff
@@ -0,0 +1,78 @@
...
+endef
+
+all: clean vendor test/server/wait test ## Set up and test
+
+node_modules: ## Run 'npm ci' if directory doesn't exist
+	$(call npm, ci)
+
+vendor: node_modules ## Install source code packages
+
+engine_status_check_url := http://localhost:8081/streamr-core/api/v1/users/me
```
teogeb, line 38: Use http://localhost/api/v1/users/me

rest-e2e-tests/Makefile
```diff
@@ -0,0 +1,78 @@
...
+		if [ $$i -eq 70 ]; then
+			echo "test/server/wait: timeout while waiting for engine to come online" 1>&2
+			exit 1
+		fi
+	done
+
+.PHONY: test
+test: vendor test/e2e test/stress ## Run tests
+
+.PHONY: test/e2e
```
teogeb, line 63: Why naming convention is different from main makefile where names are like `test-integration`?

teogeb (approved):

rest-e2e-tests/Makefile
kare: Changing url breaks the logic of the script. Nginx replies too quick while engine is down.

rest-e2e-tests/Makefile
kare: It's an experimental feature hidden behind the main Makefile. Let's see if those test target's get any use and if they are better the old way. Consider 'test/' as root 'folder' for all test related recipes and you can easily check what's available with `make test\t`.

PR #879, opened 2020-11-24
BACK-29: Support DataUnion v2 join requests

teogeb:
Checks the dataUnionVersion from product. If the version is 2, uses Java client (DU2-alpha.1) to add the member. DUv1 functionality doesn't change.


kare (approved):
LGTM

PR #880, opened 2020-11-24
DEVOP-16: Rebase Core database dumps

teogeb:
All DB migration scripts have been applied to core_dev.sql and core_test.sql database dumps (in streamr-docker-dev). We can remove the script files.


kare (approved):
LGTM

PR #882, opened 2020-11-26
BACK-103: Fix enum mappings

kare:
Grails warns about mappings on server startup
* Use defaults to suppress warnings
* Replace 'identity' with 'ordinal'
* Identity is Grails 3 feature


grails-app/domain/com/unifina/domain/Canvas.groovy
```diff
@@ -70,7 +70,7 @@ class Canvas {
 		adhoc defaultValue: false
 		runner index: 'runner_idx'
 		dashboardItems cascade: 'all-delete-orphan'
-		exampleType enumType: "identity", defaultValue: ExampleType.NOT_SET, index: 'example_type_idx'
+		exampleType enumType: "ordinal", index: "example_type_idx"
```
teogeb, line 73: Why we don't define the defaultValue here? (and 3 other places in this PR)

kare: @teogeb It doesn't do anything with enums

teogeb: How about if we set defaultValue: `ExampleType.NOT_SET.ordinal()`?

PR #890, opened 2020-12-02
BACK-113: Report correct URL when logging errors

kare:
Fallback reporter reports HTTP/1.0 for an unknown reason.


grails-app/controllers/com/unifina/controller/ErrorController.groovy
```diff
@@ -69,7 +69,11 @@ class ErrorController {
 
 		// Log internal errors
 		if (apiError.statusCode >= 500 && apiError.statusCode < 600) {
-			log.error("Unexpected error occurred on request to ${request?.getMethod()} ${request?.getRequestURL()}, returning status code 500", exception)
```
teogeb, line 72: Maybe we should keep this error message and just add another way to analyze the URL?

teogeb (approved):

PR #895, opened 2021-01-13
BACK-125: Update Web3j 4.8.3

kare:
Web3j v5.0.0 is invalid.
https://github.com/web3j/web3j/commit/b7a21ad80922fc3fc94d946b9ada903035105c17


kare:
Great, now Travis has disappeared completely! 🤬

PR #902, opened 2021-01-19
DEVOP-46: Build and push Docker image with GitHub Actions

kare:
- Push to master branch triggers workflow or temporarily with workflow dispatch
- Specify exact Node version for end to end tests. GitHub Actions `setup-node` requires exact version.
### Other
- Improve Dockerfile:
    - Default Java memory: 512m max: 1024m
    - Make a distinction between `build` and `src` in `Dockerfile`. Change builder to use `src` directory
    - Try to make separation between builder and Tomcat images/phases in `Dockerfile` more obvious
- Set Grails Tomcat plugin to `provided` scope in `BuildConfig.groovy`


kare:
paths-ignore and paths check the whole PRs files. If you modify only markdown files in one commit, but the PR contains files other than markdown then build gets triggered!

PR #914, opened 2021-02-04
Try Ubuntu 20.04 on GA workflow

kare:
Run `sed -i '' -e 's/runs-on: ubuntu-latest/runs-on: ubuntu-20.04/g' *.yaml` on GitHub Actions workflow files.


kare:
Ubuntu 20.04 works after updating streamr-docker-dev-action

PR #930, opened 2021-02-23
Convert REST e2e tests to TypeScript

teogeb:
Added also `address` field to test users (`test-utilities.ts`)

### DIFF
--- a/rest-e2e-tests/types.ts
+++ b/rest-e2e-tests/types.ts
@@ -1,0 +1,4 @@
+interface EthereumAccount { 
+    privateKey: string
+    address: string
+}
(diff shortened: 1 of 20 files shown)


rest-e2e-tests/types.ts
```diff
@@ -0,0 +1,4 @@
+interface EthereumAccount {
```
kare, line 1: Let's have one type per file. Ie. rename to `EthereumAccount.ts`

teogeb: Fixed

kare (approved):
LGTM

PR #936, opened 2021-03-12
Grant stream permissions also on DU2 join

hpihkala:


hpihkala:
Tests green on travis, but some flaky canvas stress test failed on Github Actions

PR #937, opened 2021-03-15
Update end-to-end tests to use JS-client v5.x

teogeb:
Updated test to be compatible with JS-client v5.x

Small fix to swagger documentation about permission.id


kare (approved):
LGTM

PR #955, opened 2021-04-09
NET-187: Notify storage node assignment in background

teogeb:
When stream is added to storage node or removed from it, the E&E notifies the broker by publishing event to `0xFCAd0B19bB29D4674531d6f115237E16AfCE377c/storage-node-assignments` stream. Changed that publish task to be asynchronous background operation. 

The add/remove operation doesn't need to wait for the notification task to complete. E&E can respond with a success status even if the notification fails from some reason.

### DIFF
--- a/src/java/com/unifina/service/NotifyStorageNodeTask.java
+++ b/src/java/com/unifina/service/NotifyStorageNodeTask.java
@@ -1,0 +1,54 @@
+package com.unifina.service;
+
+import com.streamr.client.StreamrClient;
+import com.unifina.domain.Stream;
+import com.unifina.domain.StreamStorageNode;
+import com.unifina.domain.EthereumAddress;
+import org.apache.log4j.Logger;
+import java.util.Map;
+import java.util.HashMap;
+
+public class NotifyStorageNodeTask extends Thread {
+
+    private static final Logger log = Logger.getLogger(NotifyStorageNodeTask.class);
+
+    public enum AssigmentEvent {
+	    STREAM_ADDED,
+	    STREAM_REMOVED
+    }
+
+	EthereumAddress storageNodeAddress;
+	String streamId;
+	AssigmentEvent eventType;
+	StreamService streamService;
+	StreamrClientService streamrClientService;
+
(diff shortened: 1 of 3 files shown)


src/java/com/unifina/service/NotifyStorageNodeTask.java
```diff
@@ -0,0 +1,54 @@
...
+        }
+	}
+
+    private Map<String,Object> createMessage() {
+		Map<String,Object> message = new HashMap();
+        message.put("event", eventType.name());
+        Map<String,Object> stream = new HashMap();
+        stream.put("id", streamId);
+        stream.put("partitions", streamService.getStream(streamId).getPartitions());
+		message.put("stream", stream);
```
kare, line 51: Check indents

src/java/com/unifina/service/NotifyStorageNodeTask.java
```diff
@@ -0,0 +1,54 @@
...
+    public enum AssigmentEvent {
+	    STREAM_ADDED,
+	    STREAM_REMOVED
+    }
+
+	EthereumAddress storageNodeAddress;
+	String streamId;
+	AssigmentEvent eventType;
+	StreamService streamService;
+	StreamrClientService streamrClientService;
```
kare, line 24: These should be final

src/java/com/unifina/service/NotifyStorageNodeTask.java
```diff
@@ -0,0 +1,54 @@
...
+
+import com.streamr.client.StreamrClient;
+import com.unifina.domain.Stream;
+import com.unifina.domain.StreamStorageNode;
+import com.unifina.domain.EthereumAddress;
+import org.apache.log4j.Logger;
+import java.util.Map;
+import java.util.HashMap;
+
+public class NotifyStorageNodeTask extends Thread {
```
kare, line 11: This class is not intended for subclassing. Lets make it final.
And its not intended for usage outside this package so lets make it package local.

kare (approved):
LGTM

PR #980, opened 2021-05-14
BACK-160: Remove dead product

kare:


grails-app/migrations/core/2021-05-14-remove-dead-product.groovy
```diff
@@ -0,0 +1,11 @@
+package core
+
+databaseChangeLog = {
```
teogeb, line 3: Maybe this migration could done by executing these statements manually in the production DB? As this change is primarily about Stream Marketplace's product data, not about the Core API application.

kare: @teogeb Good point, that actually is the plan. I just wanted to store the SQL script somewhere.  I'll apply this to production DB and close PR with a comment.

kare:
Applied to production DB.

PR #981, opened 2021-05-17
BACK-161: Delete Stream and associated storage node assignments

kare:
* Delete Stream and associated storage nodes
* Improve StorageNodes error handling
* Move Product's embedded classes to same file (as defined by Grails manual)


grails-app/services/com/unifina/service/StorageNodeService.groovy
```diff
@@ -30,13 +32,23 @@ class StorageNodeService {
...
+			StreamStorageNode instance = new StreamStorageNode(streamService.getStream(streamId), storageNodeAddress.toString())
+			try {
+				StreamStorageNode saved = instance.save(validate: true, failOnError: true)
+				NotifyStorageNodeTask task = new NotifyStorageNodeTask(
+					storageNodeAddress,
+					streamId,
+					NotifyStorageNodeTask.AssigmentEvent.STREAM_ADDED,
+					streamService,
+					streamrClientService)
+				Thread thread = new Thread(task)
```
teogeb, line 44: Could modify `NotifyStorageNodeTask` to be a`Thread` instead of `Runnable` and set the name and uncaughtExceptionHandler in `NotifyStorageNodeTask's` constructor?

rest-e2e-tests/streams-api.test.ts
```diff
@@ -369,4 +369,71 @@ describe('Streams API', () => {
...
+            assert.equal(deleted.status, 204)
+            assert.equal(deleted.size, 0)
+        })
+
+        async function assertStreamsStorageNodeCount(streamId: string, storageNodeCount: number) {
+            const nodes = await Streamr.api.v1.storagenodes
+                .findStorageNodesByStream(streamId)
+                .call()
+            if (nodes.status == 404) {
+                return 0
```
teogeb, line 407: Should compare 0 to `storageNodeCount` instead of returning the value?

rest-e2e-tests/streams-api.test.ts
```diff
@@ -369,4 +369,71 @@ describe('Streams API', () => {
...
+            await assertStreamsStorageNodeCount(stream.id, 1);
+
+            const deleted = await Streamr.api.v1.streams
+                .delete(stream.id)
+                .withAuthenticatedUser(streamOwner)
+                .call()
+            assert.equal(deleted.status, 204)
+            assert.equal(deleted.size, 0)
+
+            // assert returns 404
```
teogeb, line 435: Not valid comment anymore? (if assert throws when the expectation fails)

grails-app/services/com/unifina/service/StorageNodeService.groovy
kare: It's recommended to implement `Runnable` instead of extending `Thread`. I.e. Composition over inheritance.

rest-e2e-tests/streams-api.test.ts
kare: It's bit hacky. I couldn't find an endpoint that returns storage nodes after the stream has been deleted. So calling that endpoint returns 404 after stream has been deleted. Any ideas how to improve this? I guess the correct answer would be to query the DB directly from e2e tests, but it's a bit late for that.

harbu: > It's bit hacky. I couldn't find an endpoint that returns storage nodes after the stream has been deleted. So calling that endpoint returns 404 after stream has been deleted. Any ideas how to improve this? I guess the correct answer would be to query the DB directly from e2e tests, but it's a bit late for that.

Just a thought, maybe have this method just return a count of storage nodes? I.e.
```ts
async function getStreamsStorageNodeCount(streamId: string): Promise<number> {
    ....
}
```

Then in test something like
```ts
assert.equal(await getStreamsStorageNodeCount('stream-id'), 5)
```  
Dunno?

grails-app/services/com/unifina/service/StorageNodeService.groovy
teogeb: Just trying to avoid duplicating the logic in lines 36-51 and 60-74. Moving those lines to a helper method would also work ok. Something like 
```
instance.save(validate: true, failOnError: true)
notifyStorageNode(streamId, STREAM_ADDED)
```

rest-e2e-tests/streams-api.test.ts
kare: Once the Stream has been deleted I can't use the end point `/api/v1/streams/${streamId}/storageNodes` to find a list of storage nodes, because calling that endpoint with a non-existing `streamId` returns 404. I think I'll remove that confusing semi-"assertion".

grails-app/services/com/unifina/service/StorageNodeService.groovy
kare: It's a valid point, but I'll just smirk at it and wait for the third instance:
https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

grails-app/services/com/unifina/service/StorageNodeService.groovy
```diff
@@ -30,13 +32,23 @@ class StorageNodeService {
...
+				thread.setUncaughtExceptionHandler(new NotifyStorageNodeTask.ErrorHandler())
+				thread.setName(String.format("AddStorageNodeTask[%s,%s]", streamId, storageNodeAddress))
+				thread.start()
+				return saved
+			} catch (Exception e) {
+				log.error("error while adding a new storage node to stream", e)
```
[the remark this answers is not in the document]
kare: Good catch!

PR #986, opened 2021-05-19
BACK-167: Optimize stream id validation

teogeb:
Updated regex patterns for custom stream id validation. Earlier the nested repetition operators made the parsing extremely slow for some inputs. See http://www.regular-expressions.info/catastrophic.html. E.g. validating of a Ethereum address without a path (`0xAbcdeabCDE123456789012345678901234567890`) required ~200k steps.

The new pattern doesn't allow ENS domain name to contain underscore (as it is not allowed in the ENS spec). Otherwise the updated pattern is compatible with the earlier pattern.

### DIFF
--- a/src/groovy/com/unifina/domain/CustomStreamIDValidator.groovy
+++ b/src/groovy/com/unifina/domain/CustomStreamIDValidator.groovy
@@ -11,12 +11,21 @@ class CustomStreamIDValidator {
 		boolean isOwnedBy(String domain, User owner)
 	}
 
+	// ENS domain rules:
+	// - must contain two or more segments separated by dot
+    // - segments can contain chars a-z, A-Z, 0-9 and -
+	// https://docs.ens.domains/contract-api-reference/name-processing
+	private static final String ENS_DOMAIN_REGEX = "(?:[a-zA-Z0-9-]+\\.)+[a-zA-Z0-9-]+"
+
 	// path rules:
 	// - must start by slash
 	// - can contain chars a-z, A-Z, 0-9 and -_.
 	// - must not end with non-word character (in this case - or .)
 	// - can have segments separated by slashes (two consecutive slashes is not allowed)
-	public static final Pattern REGEX = Pattern.compile("^((?:[\\w-]+\\.?)*\\w)/(?:[\\w\\.-]+/?)*\\w\$")
+	private static final String PATH_REGEX = "(?:(?:/[\\w\\.-]+)+\\w)|(?:/\\w)"
+
+	private static final String ETHEREUM_ADDRESS_REGEX = "0x[a-fA-F0-9]{40}"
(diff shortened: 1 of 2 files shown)


src/groovy/com/unifina/domain/CustomStreamIDValidator.groovy
```diff
@@ -11,12 +11,21 @@ class CustomStreamIDValidator {
...
+
 	// path rules:
 	// - must start by slash
 	// - can contain chars a-z, A-Z, 0-9 and -_.
 	// - must not end with non-word character (in this case - or .)
 	// - can have segments separated by slashes (two consecutive slashes is not allowed)
-	public static final Pattern REGEX = Pattern.compile("^((?:[\\w-]+\\.?)*\\w)/(?:[\\w\\.-]+/?)*\\w\$")
+	private static final String PATH_REGEX = "(?:(?:/[\\w\\.-]+)+\\w)|(?:/\\w)"
+
+	private static final String ETHEREUM_ADDRESS_REGEX = "0x[a-fA-F0-9]{40}"
```
kare, line 27: `{40}` limit is good! Any chance to get similar limit to ENS_DOMAIN_REGEX?

https://github.com/ethereum/EIPs/blob/master/EIPS/eip-137.md#name-syntax

Looks like the limit is 255 chars. I think the current database field's size is a limitation. Should that be migrated?

teogeb: The ENS domain regex is mostly for catching the most typical errors early. If it looks ok, we query the ENS registry, and that query won't return OK for invalid ENS names. 

But yes, we could test that the full streamId has max 255 chars. I'll fix.

test/unit/com/unifina/domain/CustomStreamIDValidatorSpec.groovy
```diff
@@ -31,6 +31,7 @@ class CustomStreamIDValidatorSpec extends Specification {
 		"sub.my-domain.eth/foo//bar" | false
 		"foobar.eth/abc/def" | false
 		"sub.my-domain.eth" | false
+		"0xAbcdeabCDE123456789012345678901234567890" | false
```
kare, line 34: @teogeb How about adding an e2e test for this also?

src/groovy/com/unifina/domain/CustomStreamIDValidator.groovy
teogeb: Fixed

kare: 👍

test/unit/com/unifina/domain/CustomStreamIDValidatorSpec.groovy
kare: Most of the Groovy specs get throw to the trash bin when work starts on BACK-163. Probably not the case in this test case 😀

kare (approved):
LGTM

PR #992, opened 2021-06-01
fix(storageNode): change relation to one-to-many

kare:
stream has one to many storage nodes.

### DIFF
--- a/grails-app/domain/com/unifina/domain/StreamStorageNode.groovy
+++ b/grails-app/domain/com/unifina/domain/StreamStorageNode.groovy
@@ -1,24 +1,18 @@
 package com.unifina.domain
 
+import com.unifina.utils.IdGenerator
 import grails.persistence.Entity
 import groovy.transform.EqualsAndHashCode
 import groovy.transform.ToString
 
 @Entity
-@EqualsAndHashCode(includes = "streamId,storageNodeAddress")
+@EqualsAndHashCode(includes = "id,streamId,storageNodeAddress,dateCreated")
 @ToString
-class StreamStorageNode implements Serializable {
-	String streamId
+class StreamStorageNode {
+	String id
+	String streamId // TODO: remove streamId?
 	String storageNodeAddress
 	Date dateCreated
-
-	StreamStorageNode(final Stream stream, final String storageNodeAddress) {
-		Objects.requireNonNull(stream, "stream")
-		this.stream = stream
-		this.streamId = stream.getId()
-		Objects.requireNonNull(storageNodeAddress, "storageNodeAddress")
-		this.storageNodeAddress = storageNodeAddress
-	}
 
 	static belongsTo = [
 		stream: Stream,
@@ -30,10 +24,7 @@ class StreamStorageNode implements Serializable {
 	}
 
 	static mapping = {
-		id composite: [
-			'streamId',
-			'storageNodeAddress',
-		]
+		id(generator: IdGenerator.name)
 		stream(insertable: false, updateable: false)
 	}
 
--- a/grails-app/domain/com/unifina/domain/StreamStorageNode.groovy
+++ b/grails-app/domain/com/unifina/domain/StreamStorageNode.groovy
@@ -10,7 +10,7 @@ import groovy.transform.ToString
 @ToString
 class StreamStorageNode {
 	String id
-	String streamId // TODO: remove streamId?
+	String streamId
 	String storageNodeAddress
 	Date dateCreated
 
--- a/grails-app/migrations/core/2021-05-14-stream-fk-storage-node.groovy
+++ b/grails-app/migrations/core/2021-05-14-stream-fk-storage-node.groovy
@@ -1,15 +1,63 @@
 package core
 
+import com.unifina.utils.IdGenerator
+
 databaseChangeLog = {
 	changeSet(author: "kkn", id: "stream-fk-storage-node-1") {
-		addForeignKeyConstraint(
-			constraintName: "fk_stream_has_storage_nodes",
-			baseTableName: "stream_storage_node",
-			baseColumnNames: "stream_id",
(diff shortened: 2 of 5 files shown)


grails-app/domain/com/unifina/domain/StreamStorageNode.groovy
```diff
@@ -1,25 +1,19 @@
 package com.unifina.domain
 
+import com.unifina.utils.IdGenerator
 import grails.persistence.Entity
 import groovy.transform.EqualsAndHashCode
 import groovy.transform.ToString
 
 @Entity
-@EqualsAndHashCode(includes = "streamId,storageNodeAddress")
+@EqualsAndHashCode(includes = "id,streamId,storageNodeAddress,dateCreated")
```
teogeb, line 9: If `id` uniquely identifies the object, we don't need to consider `streamId`, `storageNodeAddress` or `dateCreated` in `equals()` and `hashCode`?

grails-app/services/com/unifina/service/StorageNodeService.groovy
```diff
@@ -12,25 +12,53 @@ class StorageNodeService {
...
+	StreamStorageNode findStorageNodeByAddress(Stream stream, EthereumAddress address) {
+		for (final StreamStorageNode node : stream.getStorageNodes()) {
+			if (node.getStorageNodeAddress() == address.toString()) {
+				return node
+			}
+		}
+		return null
+	}
+
+	boolean hasStorageNodeWithAddress(Stream stream, EthereumAddress address) {
```
teogeb, line 42: This method could be simplified to `findStorageNodeByAddress(s, a) != null` (or could remove the method)

grails-app/services/com/unifina/service/StorageNodeService.groovy
```diff
@@ -41,15 +69,23 @@ class StorageNodeService {
 			thread.setUncaughtExceptionHandler(new NotifyStorageNodeTask.ErrorHandler())
 			thread.setName(String.format("AddStorageNodeTask[%s,%s]", streamId, storageNodeAddress))
 			thread.start()
-			return saved
+			return node
 		} else {
+			// TODO: https://linear.app/streamr/issue/BACK-155/assign-a-stream-to-a-storage-node-when-it-has-already-been-assigned
```
teogeb, line 74: In this case the method just return the existing `StreamStorageNode` (as BACK-155 states that nothing should happen if the stream has already been assigned)

grails-app/services/com/unifina/service/StorageNodeService.groovy
kare: I've got patch coming up after this one that removes the method. It's a fix for BACK-155

grails-app/migrations/core/2021-05-14-stream-fk-storage-node.groovy
```diff
@@ -1,15 +1,42 @@
 package core
 
+import com.unifina.utils.IdGenerator
+
 databaseChangeLog = {
 	changeSet(author: "kkn", id: "stream-fk-storage-node-1") {
-		addForeignKeyConstraint(
-			constraintName: "fk_stream_has_storage_nodes",
+		sql("alter table stream_storage_node drop primary key")
+		sql("alter table stream_storage_node drop index stream_storage_node_stream_idx")
```
teogeb, line 8: Do we need to drop these indexes and then create same indexes at lines 29, 30?

grails-app/domain/com/unifina/domain/StreamStorageNode.groovy
kare: I think the contract for `equals()` and `hashCode()` is more complex than that.

If you are comparing two `StreamStorageNode`'s that have same `ID`, but other has been updated later objects should not be equal.

BTW I've started using https://jqno.nl/equalsverifier/ in new Streamr Java Client. I recommend that for all Java projects!

grails-app/services/com/unifina/service/StorageNodeService.groovy
kare: I'll fix it later. Frontend might need some notification about it.

grails-app/domain/com/unifina/domain/StreamStorageNode.groovy
teogeb: True, if we think StreamStorageNode objects as mutable. When we just create and delete items, it would be maybe be ok think items as immutable objects. But if the objects are mutable, we could just compare `streamId` and `storageNodeAddress`?

grails-app/migrations/core/2021-05-14-stream-fk-storage-node.groovy
kare: Might not be strictly necessary, but it makes the DB rebuild the index. (Granted that in a DB of our size it probably isn't required.)

grails-app/domain/com/unifina/domain/StreamStorageNode.groovy
kare: @teogeb Yes in theory, but we are not working with immutable objects here. If all fields of the class are defined with `final` then object is immutable and implementing `equals` and `hashCode` is pretty easy. I don't think that Hibernate allows `final` fields for object properties.

It might work with just `streamId` and `storageNodeAddress`, but there might be a corner case where `dateCreated` is needed.

If someone uses `StorageNode` (where only `id` is used in equals) as a key in a `Map` before `StorageNode` was saved to database `hashCode` is not going to work.

Since we really don't know how `StorageNode` class is going to be used it's safer to include all significant fields to `equals` and `hashCode`.

Note that again inheritance makes things super complex here: Use `instanceof` check in `equals` instead of `getClass()` comparison. This allows comparing classes that inherit each other and makes things a lot harder than necessary.

PR #993, opened 2021-06-01
fix(api/storage): Fixes BACK-155

kare:
When a client tries to assign storage node again (POST /streams/{id}/storageNodes) to the same storage node replies with 200 instead of 400.


kare:
@juhah @tumppi Does frontend need to be deployed at the same time?

tumppi:
> @juhah @tumppi Does frontend need to be deployed at the same time?

Should be fine without redeploying frontend 👍

PR #998, opened 2021-06-14
Improve validation error messages

teogeb:
The `ValidationException` message now includes the rejected value. The content is shorter and easier to interpret, e.g.:

```
Invalid id: 0x1234567890/foobar
Invalid partitions: 999 (max.exceeded)
Invalid category: non-existing-category-id (typeMismatch)
```


kare (approved):
LGTM

PR #1004, opened 2021-06-24
BACK-182: User is Ethereum address

kare:

PR #1007, opened 2021-07-01
BACK-182: User is Ethereum address

hpihkala:
This is #1004 reopened after reverting

* Convert integration keys to users

* Remove IntegrationKeyApiController

* Do not remove integration_key table yet

* Calculate DATA balance without IntegrationKeys

* Join DataUnion without IntegrationKey

* Remove IntegrationKey's from DataUnion*ApiController

* Fix migration image upload

* Remove IntegrationKey from StreamService

* Remove remaining IntegrationKeys


hpihkala:
@kare let's find a way to soon get back to this reverted PR that was almost ready but stalled during the holidays. Once the new website is out, the frontend guys should have some bandwidth to sync changes.

PR #1047, opened 2021-08-28
fix: slow permission query killing core-api

jtakalai:
Problem:
* MySQL cpu usage at 770%
* `show full processlist;` showing 102 processes from streamr-prod-2 in queue at all times
* those queries VERY slow (50 seconds)

Quick testing:
* get only user-specific permission: 0.2s
* get only anonymous permission (when there are none): 5s
* get user-specific OR anonymous permission: 50s

Solution: only ask MySQL for anonymous permissions if no user-specific permissions found.

Reason could be a bad index: in `TABLE permission`, there is `KEY `anonymous_idx` (`anonymous`)` but that column only has 2 possible values. So maybe not so good to index it.

### DIFF
--- a/grails-app/services/com/unifina/service/PermissionStore.groovy
+++ b/grails-app/services/com/unifina/service/PermissionStore.groovy
@@ -10,23 +10,39 @@ class PermissionStore {
 	}
 
 	List<Permission> findDirectPermissions(String resourceProp, Object resource, Permission.Operation operation, Userish userish) {
-		List<Permission> directPermissions = Permission.withCriteria {
-			eq(resourceProp, resource)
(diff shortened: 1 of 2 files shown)


hpihkala:
This changes the semantics a little bit. There could be a user that has for example:
- publish permission directly
- subscribe permission via anonymous permission

Previously the method would return both, but after this optimization it only returns the publish permission.

Anyway it's probably a corner case, so let's try it out. The permission table has become quite polluted with permissions to demo streams automatically given to new users, let's combine this code update with a database cleanup and it will probably start working smoothly again.

hpihkala (approved):

PR #1062, opened 2021-09-16
optimize findStreamsByStorageNode

teogeb:
Optimized `/api/v1/storageNodes/:streamId/streams` endpoint. 

Previously we fetched the list of `streamIds` and the fetched each stream one by one  The new implementation uses single query to fetch all required data.

Previously the endpoint query took about 14 seconds if there is 10000 streams in a storage node (in local environment). After the optimization the same operation takes about 1 second.


kare (approved):
LGTM

PR #1084, opened 2021-10-01
Remove StorageNode notification threading

harbu:
Re-opened https://github.com/streamr-dev/core-api/pull/1066 due to having to revert it for the moment being (TN3). Let's get back to this after.


kare:
Not needed anymore

PR #1085, opened 2021-10-04
Fix bad serialization of storageNodeAddress.

timoxley:
This seems to be serialized as `{storageNode: {value: string}}` instead of just `{storageNode: string}`, which breaks the broker.

### DIFF
--- a/src/groovy/com/unifina/domain/EthereumAddress.groovy
+++ b/src/groovy/com/unifina/domain/EthereumAddress.groovy
@@ -22,6 +22,7 @@ class EthereumAddress {
 		this.value = Keys.toChecksumAddress(value)
 	}
 
+	@Override
 	public String toString() {
 		return value
 	}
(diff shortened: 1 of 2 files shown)


harbu (commented):
has something changed to cause this to happen?

timoxley:
No idea why this is suddenly happening, check #dev channel

teogeb (approved):
Before the clustering feature (https://github.com/streamr-dev/network-monorepo/pull/180), which we merged yesterday, the `storageNode` value was not read from the assignment event message. 

The assignment stream was originally implemented when we only had one storage node, and it didn't contain storage node information at all. The info was added a month ago in https://github.com/streamr-dev/core-api/pull/1052, but Broker wasn't updated.

harbu:
> Before the clustering feature ([streamr-dev/network-monorepo#180](https://github.com/streamr-dev/network-monorepo/pull/180)), which we merged yesterday, the `storageNode` value was not read from the assignment event message.
> 
> The assignment stream was originally implemented when we only had one storage node, and it didn't contain storage node information at all. The info was added a month ago in #1052, but Broker wasn't updated.

Ah that explains

harbu (approved):

PR #1150, opened 2021-12-28
BACK-200: Draft

kare:
- `/api/v1` is now `/api/v2`
- Product.streams is now a Set of Stream ID's as Strings
- Migrations for Product.streams, StreamStorageNode, Stream
- Package `com.unifina` is now `com.streamr.core`


jtakalai:
DU test fails on smart contract call `deployNewDataUnion(address,uint256,address[],string)` which is incorrect because it really should have 6 arguments and not 4. That is the old function signature. Both master branch and BACK-200 branch use streamr-client version 5.5.7 where the new 6-argument ABI exists, see `rest-e2e-tests/node_modules/streamr-client/streamr-client.web.js` lines 27023 and 28912  (after running `npm ci`) which is the file I believe is executed: both the declaration and invocation are with 6 arguments.

Other occurrences of deployNewDataUnion in that directory:
* `rest-e2e-tests/node_modules/streamr-client/src/dataunion/abi.js` line 290 with 6 inputs
* `rest-e2e-tests/node_modules/streamr-client/src/dataunion/Contracts.js` line 226 with 6 arguments

On my computer, when I run that test alone (both from `master` and `BACK-200` branches) it deploys the new data union just fine:
```
rest-e2e-tests (master)$ DEBUG=Streamr* npx ts-mocha src/dataunions-api.test.ts


  DataUnions API
    PUT /api/v1/dataunions/:id/joinRequests/:id
  Streamr:StreamrClient-83245-0 new StreamrClient StreamrClient-83245-0: { version: undefined, GIT_VERSION: undefined, GIT_COMMITHASH: undefined, GIT_BRANCH: undefined } +0ms
  Streamr:DataUnion dataUnionName generated: DataUnion-1642751717046 +0ms
  Streamr:Contracts Data Union "DataUnion-1642751717046" (mainnet: 0x045D949010ed93D9126Cb1258C806e403e608F65, sidechain: 0x90922731148756D73F21A16fD739e23a6B64955F) deployed to mainnet, waiting for side-chain... +0ms
```

The mystery here then is: where does the CI get an old client version from? Or how does it end up making a 4-argument (old signature) transaction?

jtakalai:
ah, after force push, the client version is `6.0.0-alpha.21`

jtakalai:
ok, see https://github.com/streamr-dev/network-monorepo/pull/367 for the fix

PR #1247, opened 2022-06-17
refactor: load Ethereum network based on Product.chain

kare:


hpihkala:
The DU join features in the core-api are now obsolete due to moving to the new join server, closing this

